PR #14, opened 2023-09-21
Add GQ sign/verify support on PK tokens

### LINKED ISSUE #4: Add pluggability for OIDC providers
Currently the OpkClient only supports Google as the OIDC provider:

https://github.com/bastionzero/openpubkey/blob/2c4e61e93a73d3aee20a1a54e6933836400df618/parties/opkclient.go#L57-L67

This should be pluggable so that we can create implementations for other OIDC providers too.

- [ ] #5 
- [ ] #6

### LINKED ISSUE #9: Support for verifying GQ signatures in OpenPubkey tokens
Perhaps this can be done by modifying the library used to validate JWS tokens, or just by doing JWS validation ourselves.

### LINKED ISSUE #10: Support for adding GQ signatures to OpenPubkey tokens
As OpenPubkey tokens are JWS, we need to have a way of identifying the signature as being a GQ signature. The `alg` will be set to `RS256`  because the protected headers are fixed, so we will need to add an unprotected header to flag this.

jonnystoten:
Closes #9 
Closes #10 
Closes #4

- Makes opkclient more plugable for github actions
- Use bytes.Equal instead of slices for Go 1.20
- Add GQ sign/verify support on PK tokens

Also use verbose JWS format for PK tokens. @EthanHeilman I wasn't sure if the compact JWS format with `:`s was needed so I've left the code there, if it's not needed we can remove it.

### DIFF
--- a/parties/opkclient.go
+++ b/parties/opkclient.go
@@ -1,44 +1,58 @@
 package parties
 
 import (
-	"context"
 	"crypto/ecdsa"
 	"fmt"
 	"io"
 	"net/http"
-	"os/exec"
-	"runtime"
 
-	"time"
-
-	"github.com/google/uuid"
-	"github.com/lestrrat-go/jwx/v2/jwa"
-	"github.com/lestrrat-go/jwx/v2/jwk"
 	"github.com/sirupsen/logrus"
-	"github.com/zitadel/oidc/v2/pkg/client/rp"
 	"github.com/zitadel/oidc/v2/pkg/oidc"
-
-	httphelper "github.com/zitadel/oidc/v2/pkg/http"
 
 	"github.com/bastionzero/openpubkey/pktoken"
 )
 
-var (
-	key = []byte("NotASecureKey123")
-)
+// Interface for interacting with the MFA Cosigner (MFACos)
+type MFACos interface {
+	// place holder for MFA Cosigner
+	// TODO: Add MFA Cosigner
+}
 
 type OpkClient struct {
-	PktCom       []byte
-	Signer       *pktoken.Signer
-	ClientID     string
-	ClientSecret string
-	Issuer       string
-	Scopes       []string
-	RedirURIPort string
-	CallbackPath string
-	RedirectURI  string
-	MfaURI       string
-	server       *http.Server
+	PktCom      []byte
+	Signer      *pktoken.Signer
+	Op          OpenIdProvider
+	MFACosigner MFACos
+}
+
+func (o *OpkClient) OidcAuth() {
+	nonce := o.Signer.GetNonce()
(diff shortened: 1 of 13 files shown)


EthanHeilman:
> Also use verbose JWS format for PK tokens. [..] I wasn't sure if the compact JWS format with :s was needed so I've left the code there, if it's not needed we can remove it.

I'm ok with that. Some decisions need to be made of how to represent the multisignature JWS when storing as compact format in web browsers, but I'm still working through that. Let's keep the ":" delimited compact format as a placeholder.

parties/opkclient.go
```diff
@@ -49,10 +78,15 @@ func (o *OpkClient) OidcAuth() {
...
+type PublicKey interface {
+	Equal(x crypto.PublicKey) bool
+}
+
 // Interface for interacting with the OP (OpenID Provider)
 type OpenIdProvider interface {
 	RequestTokens(cicHash string, cb TokenCallback) error
-	VerifyPKToken(pktCom []byte, cosPk *ecdsa.PublicKey) error
+	VerifyPKToken(pktJSON []byte, cosPk *ecdsa.PublicKey) error
+	PublicKey(idt []byte) (PublicKey, error)
```
EthanHeilman, line 89: Oh, I really like this pattern you are using here. Adding the OP pubkey to the OpenIdProvider interface is great place to plug in different public key resolvers: JWKS URIs, JWT Ledger, TUFs, etc...

pktoken/pktoken.go
```diff
@@ -51,6 +63,119 @@ func FromCompact(pktCom []byte) (*PKToken, error) {
...
+		CicSig:  []byte(cic.Signature),
+		CosPH:   []byte(cos.Protected),
+		CosSig:  []byte(cos.Signature),
+	}
+}
+
+func (p *PKToken) ToJWS() *JWS {
+	var opSignType string
+	if p.OpSigGQ {
+		opSignType = "oidc_gq"
```
EthanHeilman, line 124: Might be worth pulling the signing type strings out and making it an enum/constant so that the string only needs to be defined in one place

EthanHeilman (approved):
Looks good to merge. I ran tests locally and created a GQ signed PK Token and verified it. 
* If you don't want to address the sign type enum comment in this PR just put a TODO in there.
* Update the Readme so show that GQ signatures are done

pktoken/pktoken.go
jonnystoten: Good call, done!

parties/opkclient.go
jonnystoten: Yep, it's a good way to have a test provider with a static key too

PR #17, opened 2023-09-25
Fix nonce validation and allow extra CIC claims

### LINKED ISSUE #12: Allow additional claims to be added the CIC
We have a use case for binding a PK token to a single signature, so it would be useful to allow extra claims to be added in the CIC before it's signed by the OP.

### LINKED ISSUE #13: Have OpkClient return the PK token rather than writing it to disk

### LINKED ISSUE #16: Nonce isn't checked on validation
We don't actually check that the nonce is a hash of the CIC when verifying a PK token. The check currently reads the nonce from the PK and then checks that the PK token contains the same nonce, and of course this will always be true. We should instead construct it from the CIC.

We have a function to compute the nonce hash already which is used to get the nonce when signing, but it suffers from the same JSON key order issue mentioned at the end of #15 - I haven't checked but wouldn't be surprised if the nonces we're generating aren't actually hashes of the CIC protected headers. We should instead hash the exact bytes that are going into the JWS. This might mean that we need to do some more manual signing of the JWS rather than leaning on the jws library.

jonnystoten:
A few things I've fixed and added while wiring the library up in one of my projects.

Closes #12
Closes #13
Closes #16

- Fix nonce validation
- Pass extra CIC claims on sign and return on verify
- Return ID token from OidcLogin func

@EthanHeilman let me know if you'd prefer this as multiple PRs!

### DIFF
--- a/parties/googleclient.go
+++ b/parties/googleclient.go
@@ -1,9 +1,11 @@
 package parties
 
 import (
+	"bytes"
 	"context"
 	"crypto/ecdsa"
 	"crypto/rsa"
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"os/exec"
@@ -100,24 +102,13 @@ func (g *GoogleOp) VerifyPKToken(pktJSON []byte, cosPk *ecdsa.PublicKey) error {
 		return err
 	}
 
-	nonce, err := pkt.GetNonce()
+	cicphJSON, err := util.Base64DecodeForJWT(pkt.CicPH)
 	if err != nil {
 		logrus.Fatalf("Error parsing PK Token: %s", err.Error())
 		return err
 	}
 
-	options := []rp.Option{
-		rp.WithVerifierOpts(rp.WithIssuedAtOffset(5*time.Second), rp.WithNonce(func(ctx context.Context) string { return string(nonce) })),
-	}
-
-	googleRP, err := rp.NewRelyingPartyOIDC(
-		g.Issuer, g.ClientID, g.ClientSecret, g.RedirectURI, g.Scopes,
-		options...)
-
-	if err != nil {
-		logrus.Fatalf("Failed to create RP to verify token: %s", err.Error())
-		return err
-	}
+	nonce := string(util.B64SHA3_256(cicphJSON))
 
 	idt := pkt.OpJWSCompact()
 	if pkt.OpSigGQ {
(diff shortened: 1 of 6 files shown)


parties/googleclient.go
```diff
@@ -64,15 +66,19 @@ func (g *GoogleOp) RequestTokens(cicHash string, cb TokenCallback) error {
...
 
 	marshalToken := func(w http.ResponseWriter, r *http.Request, tokens *oidc.Tokens[*oidc.IDTokenClaims], state string, rp rp.RelyingParty) {
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
+			chErr <- err
 			return
 		}
-		cb(tokens)
 
+		ch <- []byte(tokens.IDToken)
```
EthanHeilman, line 81: Using channels rather than a callback here has been on my mental task list for longer than I'd like to admit. Thank you for doing this, so much easier to read and think about

parties/googleclient.go
```diff
@@ -87,88 +93,124 @@ func (g *GoogleOp) RequestTokens(cicHash string, cb TokenCallback) error {
...
 
-	options := []rp.Option{
-		rp.WithVerifierOpts(rp.WithIssuedAtOffset(5*time.Second), rp.WithNonce(func(ctx context.Context) string { return string(nonce) })),
-	}
-
-	googleRP, err := rp.NewRelyingPartyOIDC(
-		g.Issuer, g.ClientID, g.ClientSecret, g.RedirectURI, g.Scopes,
-		options...)
-
+	cicphJSON, err := util.Base64DecodeForJWT(pkt.CicPH)
```
EthanHeilman, line 121: Can you think of any reason we shouldn't change the spec to hash the base64 encoding of the CIC? That way a verifier would never need to decode a CIC protected header that wasn't committed to in nonce. It would prevent someone from messing with the [unused bits in a base64 encoding.](https://developer.fastly.com/reference/vcl/functions/cryptographic/digest-base64-decode/#:~:text=Unused%20bits&text=These%20ignored%20bits%20could%20have,first%20two%20bits%20of%20A%20.) 

This also fits closer with how JWS computes the signature:

> Compute the JWS Signature in the manner defined for the
       particular algorithm being used over the JWS Signing Input
       ASCII(BASE64URL(UTF8(JWS Protected Header)) || '.' ||
       BASE64URL(JWS Payload)). 
       
https://www.rfc-editor.org/rfc/rfc7515.html#section-5.1

pktoken/signer.go
```diff
@@ -157,22 +151,32 @@ func (s *Signer) CreatePkToken(idtCom []byte) (*PKToken, error) {
...
 }
 
-func (s *Signer) CicSignature(payload []byte) ([]byte, error) {
+func (s *Signer) CicSignature(payload []byte) ([]byte, []byte, error) {
 	jwkPK := s.GetPubKey()
 	hdrs := jws.NewHeaders()
 	hdrs.Set(`upk`, jwkPK)
 	hdrs.Set(`rz`, s.rz)
 	hdrs.Set(`alg`, s.alg)
+	for k, v := range s.extraCIC {
```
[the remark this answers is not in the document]
jonnystoten: Good call, I've added this to places where we assign `extraCIC` so that it works for here and `ComputeNonce`

parties/googleclient.go
jonnystoten: I think this is a great idea! As an aside, I guess we should we have a repo to host the spec in the openpubkey org?

EthanHeilman: Yes, I think so

EthanHeilman (approved):
looks good to merge

PR #22, opened 2023-09-29
Remove println on successful verification

jonnystoten:


EthanHeilman (approved):
Now that we have channels there is no return for this println statement. In favor of removing.

Issue #15, opened 2023-09-21
GQ Signature flagging


EthanHeilman:
Reading through https://github.com/bastionzero/openpubkey/pull/14 I had the following thoughts I wanted record about possible ways to flag that a signature if GQ Signature.

Consider the following two approaches:

The unsigned header flag approach
====

We flag that a OP RSA signature has been replaced by a GQ Signature by adding a unsigned header field: `"sig_type": "oidc_gq"` Lets call 

```json
{
	"payload": "payload",
	"signatures": [{
		"protected": {"alg":"ES256","rz":"5","upk":{"crv":"P-256","kty":"EC","x":"RI","y":"Hk"}},
		"header": {
			"sig_type": "cic"
		},
		"signature": "user signature"
	}, {
		"protected": {"alg":"RS256","kid":"6f7254101f56e41cf35c9926de84a2d552b4c6f1","typ":"JWT"},
		"header": {
			"sig_type": "oidc_gq"
		},
		"signature": "GQ signature"
	}]
}
```

The altered protected header approach
====

Alternatively we could alter the protected header changing the `alg` field to `GQ256` and then signing the altered protected header and payload with the GQ signature. A verifier would then reconstruct the message the OP signed with RSA by changing the algorithm field back. It would look like:

```json
{
	"payload": "payload",
	"signatures": [{
		"protected": {"alg":"ES256","rz":"5","upk":{"crv":"P-256","kty":"EC","x":"RI","y":"Hk"}},
		"header": {
			"sig_type": "cic"
		},
		"signature": "user signature"
	}, {
		"protected": {"alg":"GQ256","kid":"6","typ":"JWT"},
		"signature": "GQ signature"
	}]
}
```

Comparison
====

**The unsigned header flag approach**

**Pros:**
* Results in a simpler implementation of the the verifier and the signer.
* None of the values which are signed by OP are altered or reserialized.

**Cons:**
* Standard OIDC validators don't know about GQ signatures or the `sig_type` header so they will likely attempt to verify the GQ signature as if it was a very large RSA signature resulting in a hard to debug failure.
* Uses an unsigned header, which means we have to worry about what happens if someone alters this header. Ideally there is no content which is not covered by a signature. This is a minor con.

**The altered protected header approach**

**Pros:**
* Standard OIDC validators will see that the `alg` is set to algorithm they don't support and fail gracefully producing a meaningful error message: Signature algorithm `GQ256` is unsupported.
[...]

PR #24, opened 2023-10-04
Rejig oidc provider interface and packages

jonnystoten:
- Add NonceClaimName and VerifyOIDCSig methods to provider interface
- Extract common stuff from VerifyPKToken methods to single function and
  remove from interface
- Add context to interface methods
- Rename parties -> oidcprovider and move CA to its own package

### DIFF
--- a/oidcprovider/oidc.go
+++ b/oidcprovider/oidc.go
@@ -1,0 +1,107 @@
+package oidcprovider
+
+import (
+	"context"
+	"crypto"
+	"crypto/rsa"
+	"encoding/json"
+	"fmt"
+
+	"github.com/lestrrat-go/jwx/v2/jwa"
+	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/openpubkey/openpubkey/pktoken"
+)
+
+const gqSecurityParameter = 256
+
+type PublicKey interface {
+	Equal(x crypto.PublicKey) bool
+}
+
+var ErrNonGQUnsupported = fmt.Errorf("non-GQ signatures are not supported")
+
+// Interface for interacting with the OP (OpenID Provider)
+type OpenIdProvider interface {
+	RequestTokens(ctx context.Context, cicHash string) ([]byte, error)
+	PublicKey(ctx context.Context, idt []byte) (PublicKey, error)
+	NonceClaimName() string
+	VerifyOIDCSig(ctx context.Context, idt []byte, expectedNonce string) error
+}
+
+func VerifyPKToken(ctx context.Context, pkt *pktoken.PKToken, provider OpenIdProvider, cosPk crypto.PublicKey) error {
+	cic, err := pkt.GetCicValues()
+	if err != nil {
+		return err
+	}
+
+	commitment, err := cic.Hash()
+	if err != nil {
+		return err
+	}
+
+	idt, err := pkt.Compact(pkt.Op)
+	if err != nil {
+		return fmt.Errorf("")
+	}
+
+	sigType, ok := pkt.ProviderSignatureType()
+	if !ok {
+		return fmt.Errorf("provider signature type missing")
+	}
+
+	switch sigType {
+	case pktoken.Gq:
+		// TODO: this needs to get the public key from a log of historic public keys based on the iat time in the token
+		pubKey, err := provider.PublicKey(ctx, idt)
+		if err != nil {
+			return fmt.Errorf("failed to get OP public key: %w", err)
+		}
+
+		rsaPubKey := pubKey.(*rsa.PublicKey)
+
+		err = pkt.VerifyGQSig(rsaPubKey, gqSecurityParameter)
+		if err != nil {
+			return fmt.Errorf("error verifying OP GQ signature on PK Token: %w", err)
+		}
+
+		var payload map[string]any
+		err = json.Unmarshal(pkt.Payload, &payload)
+		if err != nil {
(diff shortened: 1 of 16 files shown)


oidcprovider/oidc.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+var ErrNonGQUnsupported = fmt.Errorf("non-GQ signatures are not supported")
+
+// Interface for interacting with the OP (OpenID Provider)
+type OpenIdProvider interface {
+	RequestTokens(ctx context.Context, cicHash string) ([]byte, error)
+	PublicKey(ctx context.Context, idt []byte) (PublicKey, error)
+	NonceClaimName() string
+	VerifyOIDCSig(ctx context.Context, idt []byte, expectedNonce string) error
```
errordeveloper, line 32: Why not just `VerifySignature`?

oidcprovider/opkclient.go
```diff
@@ -1,8 +1,7 @@
-package parties
+package oidcprovider
```
errordeveloper, line 1: Could it be called `providers`?

oidcprovider/oidc.go
jonnystoten: It's to differentiate from the GQ signature - this is for verifying the signature from the OIDC provider where a GQ signature is not present.

To be honest, I'm thinking more and more that we should simply always use GQ signatures. We already made it an error not to use GQ signatures for Github Actions, maybe we should do the same for Google and future providers too... WDYT @EthanHeilman ?

oidcprovider/opkclient.go
lgmugnier: or `client`?

jonnystoten: Yeah that's better, or maybe `provider` (never sure whether singular or plural is better). I wanted to call this `oidc` but it was annoying to conflict with the https://github.com/zitadel/oidc library

oidcprovider/oidc.go
errordeveloper: That makes a lot of sense, maybe there could be a boolean arg or filed in opts struct to skip.

oidcprovider/opkclient.go
errordeveloper: A simple noun is better than an acronym, plural makes sense to me in this case, but it is not a big deal. `client` is okay to me, but it's super ambiguous and if we have another option, I'd go with that, which we do in this case.

lgmugnier: I would argue that technically none of the objects in this package are oidc providers, but modified oidc clients used for enacting the opk protocol against oidc providers. `oidc` seems better than `provider` here.

errordeveloper: I think it's okay to call it a provider since it's a client-side representation of a provider.
I doubt this library would ever include an implementation of a full-blown OIDC server.
In any case, I'm okay with `client`, but I think it's a bit silly to use the acronym OIDC through a library that mostly deals with OIDC.

oidcprovider/oidc.go
lgmugnier: I think it would be nice to add a new claim to the `cic` for `zkp=gq` (or something) and not allowing it to be `none` so that we can always default to `gq` for now, and not have to maintain the vanilla oidc verification, but still allow for flexibility if there are any issues we find with `gq` down the line (security-wise, resource-wise, or other).

oidcprovider/opkclient.go
lgmugnier: sorry, I know this is harping on a simple naming issue for too long but...now that I've been neck-deep in this code for a little, what do you guys think about this?
```
.
├── client
    └── opkclient.go
    └── providers
        ├── google.go
        └── github.go
```

oidcprovider/oidc.go
EthanHeilman: @jonnystoten For the workload identity I think we should require GQ signatures. In the user identity, we want allow both GQ and non-GQ.:
* GQ signatures are large and would introduce noticeable latency for say a user logging into a host over SSH or signing an API request in a browser.
* In the access use case, PK Tokens are not public so GQ signatures are less needed.

EthanHeilman: I created this issue to discuss the best way to flag that GQ signatures are being used. Please comment thoughts and ideas on that issue and then we can discuss and come to a conclusion.
https://github.com/openpubkey/openpubkey/issues/15

oidcprovider/oidc.go
```diff
@@ -0,0 +1,124 @@
...
+	Equal(x crypto.PublicKey) bool
+}
+
+var ErrNonGQUnsupported = fmt.Errorf("non-GQ signatures are not supported")
+
+// Interface for interacting with the OP (OpenID Provider)
+type OpenIdProvider interface {
+	RequestTokens(ctx context.Context, cicHash string) ([]byte, error)
+	PublicKey(ctx context.Context, idt []byte) (PublicKey, error)
+	NonceClaimName() string
```
EthanHeilman, line 31: CicHashClaim?

oidcprovider/opkclient.go
jonnystoten: Apologies for the delay getting back on this! This LGTM, will update

oidcprovider/oidc.go
EthanHeilman: I like that!

lgmugnier (approved):
Definitely a great move in the right direction!

PR #25, opened 2023-10-05
Remove IDE config

errordeveloper:
It looks like this file is unrelated to openpubkey project was checked in by accident.


jonnystoten (approved):
Thanks!

PR #30, opened 2023-10-10
Remove pktoken compaction

lgmugnier:
After a discussion with @EthanHeilman, we came to the conclusion that compaction is great, but we should remove this code for now and create a formal proposal for how we want compaction to work and look like considering there is no official standard to support multi-sig jws yet.

This PR:
1. Removes compaction functions `ToCompact` and `FromCompact`
2. Replaces that functionality with json in the x509 cert code + testing
3. Adds new `pktoken_test.go` that checks whether json is getting converted to and from a pktoken correctly


pktoken/pktoken_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+func TestPkToken(t *testing.T) {
+	TestPKTokenJSON(t)
+}
+
+func TestPKTokenJSON(t *testing.T) {
+	mockPkToken := map[string]any{
+		"payload": "eyJpc3MiOiJodHRwczovL2FjY291bnRzLmdvb2dsZS5jb20iLCJhenAiOiIxODQ5NjgxMzg5MzgtZzFmZGRsNXRnbG83bW5sYmRhazhoYnNxaGhmNzlmMzIuYXBwcy5nb29nbGV1c2VyY29udGVudC5jb20iLCJhdWQiOiIxODQ5NjgxMzg5MzgtZzFmZGRsNXRnbG83bW5sYmRhazhoYnNxaGhmNzlmMzIuYXBwcy5nb29nbGV1c2VyY29udGVudC5jb20iLCJzdWIiOiIxMDQ4NTIwMDI0NDQ3NTQxMzYyNzEiLCJlbWFpbCI6ImFub24uYXV0aG9yLmFhcmR2YXJrQGdtYWlsLmNvbSIsImVtYWlsX3ZlcmlmaWVkIjp0cnVlLCJhdF9oYXNoIjoiVmFGaGtlTE9ITXBxVWQ0RU9ZdW84ZyIsIm5vbmNlIjoiNndXMTExY25BajBlZUxzUGFDOGc5WlVkOXRDS2o1ZGNNZkt6OUZYZUFzYyIsIm5hbWUiOiJBbm9ueW1vdXMgQXV0aG9yIiwicGljdHVyZSI6Imh0dHBzOi8vbGgzLmdvb2dsZXVzZXJjb250ZW50LmNvbS9hL0FBY0hUdGRWR0Zab19aXzNoajY2ZFgzWjBHVklVUktLb2dCcGlKaDduLVhnPXM5Ni1jIiwiZ2l2ZW5fbmFtZSI6IkFub255bW91cyIsImZhbWlseV9uYW1lIjoiQXV0aG9yIiwibG9jYWxlIjoiZW4iLCJpYXQiOjE2ODUzOTU0MDEsImV4cCI6MTY4NTM5OTAwMX0",
+		"signatures": []any{
+			map[string]any{
```
EthanHeilman, line 19: One advantage of the compact format was that if something changed and need to update a test, I could just print the compact to console and copy paste it into the test. This is more readable, how easy it is print to console? Is there a go json pretty print that will include these `JSON map[string]any{`?

pktoken/pktoken_test.go
```diff
@@ -0,0 +1,60 @@
...
+				"protected": "eyJhbGciOiJFUzI1NiIsImF1dGhfdGltZSI6MTY4NTM5NTQwMCwiY3NpZCI6Imh0dHBzOi8vY29zaWduZXIuZXhhbXBsZS5jb20iLCJlaWQiOiIxIiwiZXhwIjoxNjg1NDAyNjAwLCJpYXQiOjE2ODUzOTU0MDAsImp3ayI6eyJjcnYiOiJQLTI1NiIsImt0eSI6IkVDIiwieCI6Ii01VVZVVW15YXh6Yl9UeFZpaHZ5Y2ppLUpyWmZqZ2pYZVBDdnhnQnpIcTAiLCJ5IjoiWkZIZzJYcUZ6Vk40U0V3VXJHTHM0QU8zY25LSUNzZ19hcW5EYnZrcXJ4dyJ9fQ",
+				"header": map[string]any{
+					"sig_type": "cos",
+				},
+				"signature": "P_JwaPD7VHNPq5WQADYu7EjjXoKTrU0xSLvmhDbfqr3R3VpAq2z44_r90Yl5u3zyTxvVQttJCLfXMkyooYNdbGoyC0aZrHqJYYQXGfiPGZ2xxZtP96yaEweyCw8_FI2x_-0Uc6drwnQR7AFCuLUQZZBfsKmXjxXy4X69fNkHFcZgX9cOYDwJsWKUEixSyHrSnhGPvkw0QdYS9l3tYiEtrfW6mYSVcJqKsv-bw32UG_1W4Lgg9lr_0T3xrtwfcJ35o4hntt5bi1jEzA62oXgfeBhskvFTdncCgV5kSc-gDF1-EXqoRue6QrT7qe4MOyVysR9PrXv_DwbbwnSuGeg",
+			},
+		},
+	}
+
+	mockTokenJson, err := json.Marshal(mockPkToken)
```
EthanHeilman, line 43: Any reason not to combine line 43 with line 16 so it looks like:

```
mockPkToken :=  json.Marshal(map[string]any{
    "payload": ...
     ...
})
```

Given that this is a test, I don't see any value is adding the prefix `mock` to test variables.

`pktJson, err := json.Marshal(pktMap)`

EthanHeilman (changes requested):

examples/google/example.go
```diff
@@ -119,7 +119,7 @@ func GoogleCert() {
 
 	fmt.Println("Cert skid: " + string(skid))
 
-	skidpkt, err := pktoken.FromCompact(skid)
+	skidpkt, err := pktoken.FromJSON(skid)
```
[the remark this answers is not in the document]
EthanHeilman: The more I think about this, the more I don't like either serializing the skid as compact or as JSON. It think this should just be base64 encoded.

pktoken/pktoken.go
```diff
@@ -40,60 +40,6 @@ type PKToken struct {
 	CosSig  []byte
 }
 
-func FromCompact(pktCom []byte) (*PKToken, error) {
```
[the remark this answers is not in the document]
EthanHeilman: I created this issue to capture the discussion and the Compact serialization code. I no longer object to you deleting it, assuming the other issues are solvable.
https://github.com/openpubkey/openpubkey/issues/31

cert/ca_test.go
```diff
@@ -22,10 +22,35 @@ func TestCertCreation(t *testing.T) {
...
+	mockPkTokenJson := []byte(`{
```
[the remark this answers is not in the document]
lgmugnier: I would like to voice my opinion to some of the current naming if you're trying to solidify it in a style guide.

I'm not against using stubby, acronym names but imo you have to have something that spells those out somewhere. You can't rely on someone having read the paper and having memorized every acronym straight off the bat (personally, my memory is garbage). So something like (hypothetical examples...)
```
pkt, err := pktoken.New(...)
```
makes a lot of sense because it's generally easy to make the connection that `pkt` is `pktoken`, but
```
idt, err := op.RequestToken(...)
```
is a little harder to reason about because there's no 1 for 1. Naming the variable here `idToken` gives you more information and helps explain what `RequestToken` is actually giving you. 

We also use `pkt` in function names which suffers from the same issue. It helps to make function names abundantly clear, especially since they may be read and used entirely without looking at the source code.

examples/google/example.go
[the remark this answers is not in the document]
lgmugnier: got it! I'll base64 url encode it

pktoken/pktoken_test.go
lgmugnier: nope, i used chatgpt to do this for me because it was annoying. My hope was that people would be able to copy this object and paste in their own values or we would have a programatic way to generate these mock tokens, hopefully the latter.

pktoken/pktoken_test.go
lgmugnier: nope!

cert/ca_test.go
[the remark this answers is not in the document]
lgmugnier: also, ngl, I'm not a fan of `pktCom` because I had to read your comment in one of the files in order to undestand it. I know it's jws-standard but I don't think we lose a lot by calling it `pktCompact` but that just my opinion and it's not strong enough to argue for changing the code or anything.

pktoken/pktoken_test.go
EthanHeilman: can you use the json format you used in catest.go

cert/ca.go
```diff
@@ -109,7 +110,7 @@ func PktTox509(pktCom []byte, caBytes []byte, caPkSk *ecdsa.PrivateKey, required
 			Critical: false,
 			Value:    []byte(oidcIssuer),
 		}},
-		SubjectKeyId: []byte(pktCom),
+		SubjectKeyId: []byte(base64.URLEncoding.EncodeToString(pktJson)),
```
EthanHeilman, line 113: I was thinking about your argument that everything should be base64 URI encoded. I want to have a common place where all base64 encoding happens so that all the code is using the same encoder and we can switch it easily if we need to.

Could you which this to use `util.Base64EncodeForJWT`
https://github.com/openpubkey/openpubkey/blob/main/util/base64.go#L15C6-L15C24

pktoken/pktoken_test.go
EthanHeilman: As long as they are consistent, I am good with this.

EthanHeilman (changes requested):
Switch to the base64 util function, everything looks good.

EthanHeilman (approved):
LGTM

PR #38, opened 2023-10-13
Run tests on PRs

lgmugnier:
Before we were checking go code formatting and running `go vet` but now that we have more unit tests (and are adding more), we should make sure they pass on all PRs.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   gofmt:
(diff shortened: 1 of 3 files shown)


EthanHeilman (changes requested):
This is awesome, just mark in the readme that this is done and I'll approve
https://github.com/openpubkey/openpubkey/blob/3af45864350fdd3548b79fd1482fd06ed5e32f23/README.md?plain=1#L130

PR #39, opened 2023-10-13
Add CIC object and more tests

lgmugnier:
The main goals of this PR were to:
- improve our tests, specifically surrounding the `PkToken`.
- reduce our reliance on writing to/reading from disk. 
- distinguishing better between examples and tests.
- match the architecture more closely with the paper, so that it's easier to follow along with

The core test cases:
1. Creating a new PK Token (existing test)
2. Using the PK Token to sign a message (partial existing test)
3. Verifying that signed message using the PK Token (partial existing test)

### Creating a new PK Token
Previously, the creator of our `PkToken` struct lived in the `Signer`. So, in addition to a more traditional `pktoken.New()` function, I added a new object called the `clientinstance.Claims` (to help first-time readers spell out `cic`). This new object is responsible for creating a valid `cic` and generating the nonce.

Now, our pktoken initialization can more follow the steps of the paper in its instantiation code from the perspective of the client.
```golang
// Step 1: User generates key pair in the form of a crypto.Signer and JWK-formatted public key
signingKey, err := util.GenKeyPair(jwa.ES256)// returns object that implements crypto.Signer interface
jwkKey, err := jwk.PublicKeyOf(signingKey)
jwkKey.Set(jwk.AlgorithmKey, jwa.ES256)

// Step 2: User uses key pair to generate client instance claims
cic, err := clientinstance.NewClaims(jwk, map[string]any{})
// Step 3: Define our nonce as a commitment to our client instance claims
nonce, err := cic.Commitment()
// Step 4: Use that nonce to complete the OIDC flow and get an ID token from the provider
idToken, err := provider.RequestTokens(nonce)
// Step 5: User signs the payload from the ID token
cicToken, err := cic.Sign(signingKey, jwk.Algorithm(), idToken)
// Step 6: Combine our ID token and our signature over the cic to create our PK Token
pkt, err := pktoken.New(idToken, cicToken)
```

I made the deliberate choice to have `pktoken.New` accept separate tokens instead of just the parts we care about. This kind of structure:
- makes the `PkToken` a bit easier to integrate with other jws-type libraries.
- makes the function highly extensible. For example, if we wanted to add the ability to add multiple cosigners, we could extend the function definition `pktoken.New(idToken, pkToken string, cosTokens ...string)`.
- gives us some good storage options! If we think of the `PkToken` as a collection of jwts, we can take advantage of their compact forms.

### Signing and Verifying Messages
In order to test signing and verifying a message, I added two new functions to `PkToken`. Before, we didn't have the OpenPubKey Signed Message (OSM) so I added these two functions to build and verify those types of messages, according to the paper. We did have a sort of osm `Verify()` function before, but it wasn't being used anywhere.
```golang
NewSignedMessage(content []byte, signer crypto.Signer) ([]byte, error)
VerifySignedMessage(osm []byte) ([]byte, error)
```
I also believe these signatures would be more in line with how we would want them to be used.

### Being Intentional about Examples and Tests
The existing `example.go` code relied a lot on reading from disk. In an effort to minimize this, in addition to minimizing the amount of code that we have yet don't support, I decided to remove a lot of the code in a similar style to how it had been done before. I do think this can be added back in but seemed out of scope for this PR.

I also thought it made a lot of sense to keep this in example because it requires user interaction. So, this can be a helpful tool for manual testing although it won't be possible to automate as a unit test.

### Random Small Improvements?
- Our cic hash commitment is now base64 url encoded instead of standard.
- Added the JWK key algorithm to our JWK key when we generate key pairs and enforce that when creating a new cic.
- Tweaked previous `GenKeyPair()` function to accept `jwa.KeyAlgorithm` and return `crypto.Signer`. 
- I use the `crypto.Signer` interface more heavily here and that's mostly because that's what the jws library we're using also does.


util/files.go
```diff
@@ -75,41 +75,6 @@ func ReadCertFile(fpath string) (*x509.Certificate, error) {
 	return cert, nil
 }
 
-func SecretKeyFromBytes(pemBytes []byte) (*ecdsa.PrivateKey, error) {
-	block, _ := pem.Decode(pemBytes)
-	return x509.ParseECPrivateKey(block.Bytes)
-}
-
-func X509PublicKeyBytesFromJWK(upkjwk jwk.Key) ([]byte, error) {
```
EthanHeilman, line 83: I'm going to be using this writing to disk code fairly frequently. While I agree that we should be able to test without writing to disk, we should still keep the writing to disk functionality.

pktoken/pktoken_test.go
```diff
@@ -56,3 +77,58 @@ func TestPKTokenJSON(t *testing.T) {
...
+		return nil, err
+	}
+
+	// Sign the token with the secret key
+	signedToken, err := jwt.Sign(token, jwt.WithKey(alg, signingKey))
+	if err != nil {
+		return nil, err
+	}
+
+	return signedToken, nil
```
EthanHeilman, line 133: It would be nice to have some unittests that use known good ID Tokens from google. That way if our integration tests with Google start failing we have a past ID Token that we know worked and we can see what the failure is. Also really help from an understanding point of view to be able to point to a real ID Token from a particular OP.

examples/google/example.go
```diff
@@ -1,176 +1,35 @@
...
-		return
-	}
-	fmt.Println("Cert received: " + string(certBytes))
-
-	block, _ := pem.Decode(certBytes)
-	cert, err := x509.ParseCertificate(block.Bytes)
-	if err != nil {
-		fmt.Printf("Error failed to parse cert: %s", err.Error())
-		return
-	}
```
EthanHeilman, line 116: This code is useful, don't delete. It will be changed and updated soon

pktoken/pktoken.go
```diff
@@ -357,26 +384,18 @@ func (p *PKToken) VerifyCosSig(cosPk jwk.Key, alg jwa.KeyAlgorithm) error {
 	return nil
 }
 
-func (p *PKToken) Verify(msg string, sig []byte) error {
-	jwsSig, err := jws.Parse(sig)
+func (p *PKToken) Hash() (string, error) {
+	// get pktoken bytes
+	pktJson, err := p.ToJSON()
```
EthanHeilman, line 389: I worry that `pkt-->pktJson` introduces serialization differences. Ideally we would represent all content that has been signed in a compact format such that all signatures can be checked without deserialization.

pktoken/pktoken.go
```diff
@@ -357,26 +384,18 @@ func (p *PKToken) VerifyCosSig(cosPk jwk.Key, alg jwa.KeyAlgorithm) error {
...
-	}
-
-	alg, _, upk, err := p.GetCicValues()
-	if err != nil {
-		return err
+		return "", err
 	}
 
-	_, err = jws.Verify(sig, jws.WithKey(alg, upk))
+	hasher := sha256.New()
```
EthanHeilman, line 394: It isn't written down anywhere, but OpenPubkey should use SHA-3 except when required to use SHA-256 by the JWS standard.

pktoken/pktoken.go
```diff
@@ -357,26 +384,18 @@ func (p *PKToken) VerifyCosSig(cosPk jwk.Key, alg jwa.KeyAlgorithm) error {
...
+	_, err = hasher.Write(pktJson)
 	if err != nil {
-		return err
+		return "", err
 	}
-
-	// verified
-	return nil
+	hash := hasher.Sum(nil)
+	return string(util.Base64EncodeForJWT(hash)), nil
```
EthanHeilman, line 400: hexstring rather than base64. Users can then compare the values without deserializing.

EthanHeilman (changes requested):

parties/opkclient.go
```diff
@@ -23,49 +25,68 @@ type MFACos interface {
...
+	UserPublicKey jwk.Key // Requires "alg" header to be set
+	Gq            bool
+	Op            OpenIdProvider
+	MFACosigner   MFACos
 }
 
 func (o *OpkClient) OidcAuth() ([]byte, error) {
-	nonce, err := o.Signer.GetNonce()
+	// Use provided public key to generate client instance claims
+	cic, err := clientinstance.NewClaims(o.UserPublicKey, map[string]any{})
```
jonnystoten, line 38: The extra CIC claims here need to come from somewhere, could be a parameter to this function or a field on the `OpkClient` struct.

pktoken/clientinstance/claims.go
```diff
@@ -0,0 +1,113 @@
...
+
+	return &Claims{
+		protected: claims,
+	}, nil
+}
+
+// Returns a hash of all client instance claims which includes a random value
+func (c *Claims) Commitment() (string, error) {
+	// LUCIE: Do we need to sort to maintain ordering???
+	buf, err := json.Marshal(c.protected)
```
jonnystoten, line 54: I think it's OK as long as it's deterministic so that it's serialized in the same way when creating the token. I think it is as `json.Marshal` sorts the keys

pktoken/pktoken.go
lgmugnier: I think you made this same comment on my previous PR, but the conclusion was that you can do so with either.

pktoken/pktoken_test.go
lgmugnier: I would prefer to keep the provider-specific tests more with the provider and keep these tests a lot more agnostic. Essentially, I don't disagree but I don't think this is the right PR or place for that kind of test.

pktoken/pktoken.go
lgmugnier: 8954470 is my short term answer to this. I've added a `raw` field to our pktoken so that when we are unmarshaled from an existing json, we use whatever that was and if we're a new pktoken then we'll marshal ourselves. I personally think we're still in a stage where it's more important to match the paper (since this is essentially touted as a reference implementation) than something universally built for all languages. Hopefully, by that time we will have a new plan for compaction and we can go that route.

util/files.go
lgmugnier: e701e1c

examples/google/example.go
lgmugnier: This is pending getting the functionality fixed on main

pktoken/clientinstance/claims.go
lgmugnier: yes sweet. Just tested it out and confirmed. Thanks!

parties/opkclient.go
```diff
@@ -23,49 +25,77 @@ type MFACos interface {
...
+	alg jwa.KeyAlgorithm,
+	extraClaims map[string]any,
+	signGQ bool,
+) ([]byte, error) {
+	// Use our signing key to generate a JWK key with the alg header set
+	jwkKey, err := jwk.PublicKeyOf(signer)
+	if err != nil {
+		return nil, err
+	}
+	jwkKey.Set(jwk.AlgorithmKey, alg)
```
jonnystoten, line 44: Do we need to check that `alg` matches the `signer`? Or can we infer `alg` from `signer`? I'm wondering what would happen if I provided an ECDSA key with `alg` set to "RS256".

pktoken/osm.go
```diff
@@ -0,0 +1,88 @@
...
+		return nil, fmt.Errorf("expected only one signature on jwt, received %d", len(message.Signatures()))
+	}
+	protected := message.Signatures()[0].ProtectedHeaders()
+
+	// Verify typ header matches expected "osm" value
+	if typ, ok := protected.Get("typ"); !ok {
+		return nil, fmt.Errorf("missing required header `typ`")
+	} else if typ != "osm" {
+		return nil, fmt.Errorf(`incorrect "typ" header, expected "osm" but recieved %s`, typ)
+	}
```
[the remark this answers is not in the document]
lgmugnier: sounds perfectly good to me. I can get self conscious about how long my code is but I don't have a preference otherwise.

I also did it for the following check statements ed9493d

parties/opkclient.go
lgmugnier: I did look into this because I agree that would just be incredibly nice, but I couldn't figure out any way to do so. We could check that the `alg` is of the same type as the `signer`, but even then we'd probably do so by type checking against the golang standard library which would probably be overly restrictive for anyone using a different library that implements `crypto.Signer`.

The error message I got from trying to sign with `alg=ES256` using an RSA key was:
```
failed to generate signature for signer #0 (alg=ES256): failed to sign payload: failed to unmarshal ASN1 encoded signature: asn1: structure error: length too large
```
So....not very helpful. Probably partly because the `lestrrat-go/jwx` library doesn't type-check its keys.

I actually went back and forth on this in my implementation and here's my logic:
It's hard to be helpful here because either we can have a coupled interface where we're in charge of generating, storing, and parsing the key and alg information which makes us much less flexible against new or different signing protocols, OR we leave this as uncoupled interfaces and there is some definite room for user error that will be frustrating to debug but we are more flexible. I thought flexibility was important enough to go with the latter.

jonnystoten (approved):
LGTM! :tada:

parties/googleclient.go
```diff
@@ -112,7 +112,7 @@ func (g *GoogleOp) RequestTokens(cicHash string) ([]byte, error) {
 	}
 }
 
-func (g *GoogleOp) VerifyPKToken(pktJSON []byte, cosPk *ecdsa.PublicKey) (map[string]any, error) {
+func (g *GoogleOp) VerifyPKToken(pktJSON []byte, cosPk crypto.Signer) (map[string]any, error) {
```
EthanHeilman, line 115: since cosPK is a public key, you should use `crypto.PublicKey` here rather than `crypto.Signer`

https://pkg.go.dev/crypto#PublicKey

parties/opkclient.go
```diff
@@ -82,17 +112,21 @@ type PublicKey interface {
 // Interface for interacting with the OP (OpenID Provider)
 type OpenIdProvider interface {
 	RequestTokens(cicHash string) ([]byte, error)
-	VerifyPKToken(pktJSON []byte, cosPk *ecdsa.PublicKey) (map[string]any, error)
+	VerifyPKToken(pktJSON []byte, cosPk crypto.Signer) (map[string]any, error)
```
EthanHeilman, line 115: Should be crypto.PublicKey

examples/google/example.go
lgmugnier: adressed!

parties/googleclient.go
lgmugnier: 436ad42

pktoken/pktoken.go
EthanHeilman: DIdn't you end up doing hex tho? We should do hex (all caps) to signal "do not deserialize,  no highly esteemed deed is commemorated here,  this string is best shunned and left unserialized"

examples/google/example.go
```diff
@@ -94,53 +99,47 @@ func login(outputDir string, alg jwa.KeyAlgorithm, signGQ bool) error {
...
-	msgHash := sha256.New()
-	_, err = msgHash.Write([]byte(message))
-	if err != nil {
-		return err
-	}
-	msgHashSum := msgHash.Sum(nil)
-
-	rawSigma, err := signer.Pksk.Sign(rand.Reader, msgHashSum, crypto.SHA256)
+	msgHashSum := sha3.Sum256([]byte(message))
+	rawSigma, err := signer.Sign(rand.Reader, msgHashSum[:], crypto.SHA256)
```
EthanHeilman, line 117: This is my silly variable name, since you are touching this code, can you erase my naming sin here and change this to `sig`

examples/google/example.go
```diff
@@ -94,53 +99,47 @@ func login(outputDir string, alg jwa.KeyAlgorithm, signGQ bool) error {
...
 	if err != nil {
 		return err
 	}
 
 	fmt.Println("Signed Message:", message)
 	fmt.Println("Praise Sigma:", base64.StdEncoding.EncodeToString(rawSigma))
-	fmt.Println("Hash:", hex.EncodeToString(msgHashSum))
+	fmt.Println("Hash:", hex.EncodeToString(msgHashSum[:]))
```
EthanHeilman, line 124: Is the `[:]` needed? Does this fix a bug in my code?

parties/opkclient.go
```diff
@@ -21,55 +24,87 @@ type MFACos interface {
...
+	Pkt         *pktoken.PKToken
 	Op          OpenIdProvider
 	MFACosigner MFACos
 }
 
-func (o *OpkClient) OidcAuth() ([]byte, error) {
-	nonce, err := o.Signer.GetNonce()
+func (o *OpkClient) OidcAuth(
+	signer crypto.Signer,
+	alg jwa.KeyAlgorithm,
```
EthanHeilman, line 34: could we put the alg inside the signer? Or is there something special about `jws` algorithms that the crypto.Signer doesn't support?

If we can't combine them can we add a sanity check so someone doesn't accidentally put an alg that doesn't match the crypto.Signer's algorithm?

parties/opkclient.go
```diff
@@ -21,55 +24,87 @@ type MFACos interface {
 }
 
 type OpkClient struct {
-	PktJson     []byte
-	Signer      *pktoken.Signer
+	Pkt         *pktoken.PKToken
```
EthanHeilman, line 27: It seems strange to have a `pkt` as part of the OpkClient object, but the signer which uses that pkt isn't part of the struct.

Why not have both the signer and the pktoken as part of the struct or exclude both?

parties/opkclient.go
lgmugnier: @EthanHeilman

parties/opkclient.go
EthanHeilman: Hmmm, I see the resolved comment below. Is there a jwk signer object that we could use?

examples/google/example.go
lgmugnier: No, I'm just using a different function here which result in a `[32]byte` instead of a `[]byte` so I convert it to a byte slice by adding that

lgmugnier: Johnny pointed out the exact same [issue](https://github.com/openpubkey/openpubkey/pull/39#discussion_r1367172012), I explained my reasoning here and tagged you in the conversation. He originally resolved it, but I unresolved so you could see.

parties/opkclient.go
lgmugnier: Johnny pointed out the exact same https://github.com/openpubkey/openpubkey/pull/39#discussion_r1367172012, I explained my reasoning here and tagged you in the conversation. He originally resolved it, but I unresolved so you could see.

parties/opkclient.go
lgmugnier: I kept the pkt token included because we use it for the MFA cosigner stuff later but we never use the signer again. This might change in the future but I'm just building to what's there right now

EthanHeilman (approved):
LGTM

EthanHeilman (approved):
LGTM

PR #41, opened 2023-10-16
Fix example.go

lgmugnier:
This PR fixes example.go, in doing that is also:

1. I removed the following commands: `cagen`, `ca`, and `sss`
2. Did some cleanup work converting `pktCom` -> `pktJson` since we no longer support a multi-sig compact version
3. Made sure we were setting our values correctly in our `Signer`
4. Saving our login creds after login so that we could retrieve them for other commands
5. Moved `main()` to the top of the file
6. Move everything from `config.go` into `example.go` so that we can execute without having to build every time

## How to Test
```shell
go run example.go login
go run example.go sign
go run example.go cert
```


examples/google/example.go
```diff
@@ -40,52 +81,64 @@ func GoogleSign() {
...
-	msg, err := os.ReadFile("hunter.txt")
+	pktJson, err := client.OidcAuth()
 	if err != nil {
-		panic(err)
+		return err
+	}
+
+	// Pretty print our json token
+	var prettyJSON bytes.Buffer
+	if err := json.Indent(&prettyJSON, pktJson, "", "  "); err != nil {
```
EthanHeilman, line 94: Excited to steal this so i pretty print json everywhere.

examples/google/example.go
```diff
@@ -40,52 +81,64 @@ func GoogleSign() {
...
+	_, err = msgHash.Write([]byte(message))
 	if err != nil {
-		panic(err)
+		return err
 	}
 	msgHashSum := msgHash.Sum(nil)
-	hashhex := hex.EncodeToString(msgHashSum)
 
-	rawSigma, err := client.Signer.Pksk.Sign(rand.Reader, msgHashSum, crypto.SHA256)
+	rawSigma, err := signer.Pksk.Sign(rand.Reader, msgHashSum, crypto.SHA256)
```
EthanHeilman, line 117: I know this is my code, but rereading this I am asking myself why I choose to hash the message? Shouldn't Sign just hash the message?

EthanHeilman (approved):
Thanks for cleaning up my google example code. LGTM

examples/google/example.go
lgmugnier: So, interestingly enough, it looks like we're supposed to hash so that the function can then hash it again?

> Sign signs a hash (which should be the result of hashing a larger message) using the private key, priv.

jonnystoten: Sign doesn't hash, you have to hash it yourself and pass the digest. I'm not sure why you need to tell it which hashing function you used, maybe some signing algorithms behave differently depending on the hashing function? ecdsa just ignores the parameter: https://pkg.go.dev/crypto/ecdsa#PrivateKey.Sign

PR #42, opened 2023-10-18
GQ Benchmarking

lgmugnier:
`BenchmarkSigning` benchmarks GQ signing a Json Web Token (JWT), `BenchmarkVerifying` benchmarks GQ verifying a Json Web Token (JWT) both with a security parameter of 256. After 1056 and 1094 iterations of creating a GQ signature or verifying a GQ signature, respectively, the results were that they both take about 11 milliseconds. This is roughly equivalent to RSA signing and verifying durations, at least on the machine on which the below values were taken.

```shell
% go test -benchmem -bench . 
goos: darwin
goarch: arm64
pkg: github.com/openpubkey/openpubkey/gq
BenchmarkSigning-10                 1056           1139994 ns/op          236609 B/op        438 allocs/op
BenchmarkVerifying-10               1094           1091559 ns/op          199087 B/op        368 allocs/op
PASS
ok      github.com/openpubkey/openpubkey/gq     271.329s
```


gq/benchmark_test.go
```diff
@@ -0,0 +1,93 @@
...
+			b.Fatal(fmt.Errorf("Failed to verify signature!"))
+		}
+	}
+
+	// Avoid compiler optimisations eliminating the function under test and artificially lowering the run time of the benchmark
+	// ref: https://dave.cheney.net/2013/06/30/how-to-write-benchmarks-in-go
+	boolResult = ok
+}
+
+func generateTestMatrix(n, privateSize, messageSize int) ([]testTuple, error) {
```
jonnystoten, line 76: `privateSize` and `messageSize` are unused

jonnystoten (approved):
Nice one, this LGTM once the unused params are removed

EthanHeilman (approved):
LGTM

PR #44, opened 2023-10-20
Bug fix when assigning gq signature to pktoken

lgmugnier:
I introduced this bug in #28. `VerifyJWT()` differed from the existing code in that it returned an entire token instead of just the signature. Fixed that plus a small issue that we were printing an error when VerifyJWT returns a bool so error was always nil.


EthanHeilman (approved):
LGTM

PR #45, opened 2023-10-21
Use jws signatures

lgmugnier:
The goal of this PR was to improve quality of life a little bit in the PK Token. I noticed we were rolling a lot of our own stuff and for now, I think it makes it simpler and more robust to rely on the existing `jws` library. 

The functionality remains the same, however there were some condensations, removals, and additions. Here's what's changed:

### More JWS Signatures!
The core change of this PR is the new PKToken object, we used to keep track of headers and signatures separately but I think it makes it a lot more flexible to store them as `*jws.Signature`s. 

I created an alias `type Signature = jws.Signature` with the intent that we can always make our own object later but since we're using it as a functional arg, I think it's always best practice not to rely on the object signatures of an underlying library that isn't in go standard.
```golang
type Signature = jws.Signature

type PKToken struct {
	raw []byte // the original, raw representation of the object

	Payload []byte     // decoded payload
	Op      *Signature // Provider Signature
	Cic     *Signature // Client Signature
}
```

### More Files in Tree
This one might be controversion, but to me the `PKToken` is the powerhouse of our repo. I imagine its gonna get big and so we should try to keep it as organized as possible. What made the most sense to me is to create different files, each for a type of signature: `cos.go`, `cic.go`, and `gq.go` which hold the functions for signing and verifying each type of signature plus any constants or structs that might only apply to each. This frees up our main `pktoken.go` to only contain general functions. This might change in the future but since the goal was trying to improve QOL here, it made it significantly easier to find everything once I'd made this split and I think it lends itself well to the current architecture and functionality of our `PKToken`.
```
└── pktoken
    ├── cic.go
    ├── cos.go
    ├── gq.go
    ├── osm.go
    ├── pktoken.go
    └── pktoken_test.go
```

### New Signature Type
I collected our previous constants into a `SignatureType`. This will help us keep it nice and tight.
```golang
type SignatureType string

const (
	Oidc SignatureType = "oidc"
	Gq   SignatureType = "oidc_gq"
	Cic  SignatureType = "cic"
	Cos  SignatureType = "cos"
)
```

### Custom JSON Marshaler and UnMarshaler

The biggest change is in how we're converting to and from JSON. Before we had a few functions with supporting object etc.. I thought this would be a perfect place to rely on the `jws` library and some go magic. I've created a custom marshaler: `func (p *PKToken) MarshalJSON() ([]byte, error)`, and unmarshaler: `func (p *PKToken) UnmarshalJSON(data []byte) error`, for our object so that when we make regular json calls, the appropriate one will be used.
```golang
var pkt *pktoken.PKToken
if err := json.Unmarshal(pktJson, &pkt); err != nil {
	return err
}
```
I also added in some more formatting restriction checks for the PKToken based on what we were doing before but more verbose.

### OpSigGQ Removed
We used to have this struct variable called `OpSigGQ` which essentially said whether or not our `Op` signature was GQ or not. I didn't find this to be very extensible and I thought we could accomplish this with a simple function call for now. Here is the new function in action.
```golang
sigType, ok := pkt.ProviderSignatureType()
if !ok {
	return fmt.Errorf("provider signature type missing")
}

if sigType == pktoken.Gq {
    ...
}
```

### Condensations

1. Before, we had a function per signature type to generate the compact form, now:
```golang
func (p *PKToken) Compact(sig *jws.Signature) ([]byte, error)
```
3. Before, we had several functions to add signatures to our PkToken, now:
```golang
func (p *PKToken) AddSignature(token []byte, sigType SignatureType) error
```

### Establishing New Payload Parsing Pattern
I thought it would be healthier to start using an actual pattern for unmarshaling the payload which is stored as its raw, decoded bytes in our `PKToken`. To me, this pattern makes the most sense because it takes care of key-ing the map and parsing the values at the same time with minimal code.
```golang
var payload struct {
	Issuer   string `json:"iss"`
	Audience string `json:"aud"`
	Email    string `json:"sub"`
}
if err := json.Unmarshal(pkt.Payload, &payload); err != nil {
	return nil, err
}
```

### Parsing the CIC
Before, we were parsing our client instance claims (cic) in the pktoken and returning them. I wanted to keep all the knowledge of the construction and special headers in one place, so I've created a new function: `clientinstance.ParseClaims()`, that allows you to hydrate our claims and then interact with the object directly, getting any variable you might need like `UserPublicKey()` or calculating the commitment in exactly the same way and place it was calculated the first time around. 

### Misc

1. I had to rejig some testing to get it all working correctly and reliably


cert/ca.go
```diff
@@ -78,22 +79,26 @@ func PktTox509(pktJson []byte, caBytes []byte, caPkSk *ecdsa.PrivateKey, require
 	// 	return nil, err
 	// }
 
-	iss, aud, email, err := pkt.GetClaims()
-	if err != nil {
+	var payload struct {
+		Issuer   string   `json:"iss"`
+		Audience []string `json:"aud"`
+		Email    string   `json:"sub"`
```
EthanHeilman, line 85: The `sub` claim is not the Email, `email` is a separate claim.

The required claims for ID Tokens are `iss`, `aud` , `sub` , `exp`, `iat`. We should not deserialize non-required claims unless a particular OP aware deserializer is being used.

```
iss
REQUIRED. Issuer Identifier for the Issuer of the response. The iss value is a case sensitive URL using the https scheme that contains scheme, host, and optionally, port number and path components and no query or fragment components.
sub
REQUIRED. Subject Identifier. A locally unique and never reassigned identifier within the Issuer for the End-User, which is intended to be consumed by the Client, e.g., 24400320 or AItOawmwtWwcT0k51BayewNvutrJUqsvl6qs7A4. It MUST NOT exceed 255 ASCII characters in length. The sub value is a case sensitive string.
aud
REQUIRED. Audience(s) that this ID Token is intended for. It MUST contain the OAuth 2.0 client_id of the Relying Party as an audience value. It MAY also contain identifiers for other audiences. In the general case, the aud value is an array of case sensitive strings. In the common special case when there is one audience, the aud value MAY be a single case sensitive string.
exp
REQUIRED. Expiration time on or after which the ID Token MUST NOT be accepted for processing. The processing of this parameter requires that the current date/time MUST be before the expiration date/time listed in the value. Implementers MAY provide for some small leeway, usually no more than a few minutes, to account for clock skew. Its value is a JSON number representing the number of seconds from 1970-01-01T0:0:0Z as measured in UTC until the date/time. See [RFC 3339](https://openid.net/specs/openid-connect-core-1_0.html#RFC3339) [RFC3339] for details regarding date/times in general and UTC in particular.
iat
REQUIRED. Time at which the JWT was issued. Its value is a JSON number representing the number of seconds from 1970-01-01T0:0:0Z as measured in UTC until the date/time.
``` https://openid.net/specs/openid-connect-core-1_0.html#IDToken



Zitadel has a neat way of handling ID Token claim deserialization using omitempty. Not sure if it is a good idea or not, but it is cool:
```
type TokenClaims struct {
	Issuer                              string   `json:"iss,omitempty"`
	Subject                             string   `json:"sub,omitempty"`
``
https://github.com/zitadel/oidc/blob/main/pkg/oidc/token.go#L36

cert/ca.go
```diff
@@ -78,22 +79,26 @@ func PktTox509(pktJson []byte, caBytes []byte, caPkSk *ecdsa.PrivateKey, require
...
 	}
 
 	caTemplate, err := x509.ParseCertificate(caBytes)
 	if err != nil {
 		return nil, err
 	}
 
-	subject := string(email)
-	oidcIssuer := iss
+	subject := payload.Email
```
EthanHeilman, line 100: email != subject

examples/google/example.go
```diff
@@ -178,8 +171,8 @@ func googleCert(outputDir string, alg jwa.KeyAlgorithm, signGq bool) error {
 		return fmt.Errorf("malformatted skid: %w", err)
 	}
 
-	skidpkt, err := pktoken.FromJSON(skidDecoded)
-	if err != nil {
+	var skidpkt *pktoken.PKToken
+	if err := json.Unmarshal(skidDecoded, skidpkt); err != nil {
```
EthanHeilman, line 175: Is this handle the error case in which the skidpkt is nil? I assume so right?

parties/ca.go
```diff
@@ -189,22 +188,26 @@ func (a *Ca) PktTox509(pktCom []byte, caBytes []byte) ([]byte, error) {
 	// 	return nil, err
 	// }
 
-	iss, aud, email, err := pkt.GetClaims()
-	if err != nil {
+	var payload struct {
+		Issuer   string `json:"iss"`
+		Audience string `json:"aud"`
+		Email    string `json:"sub"`
```
EthanHeilman, line 194: email != sub

parties/githubclient.go
```diff
@@ -149,77 +147,69 @@ func (g *GithubOp) RequestTokens(cicHash string) ([]byte, error) {
...
 
-	cicphJSON, err := util.Base64DecodeForJWT(pkt.CicPH)
+	cic, err := pkt.GetCicValues()
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	expectedAud := string(util.B64SHA3_256(cicphJSON))
+	commitment, err := cic.Commitment()
```
EthanHeilman, line 165: what about `cicHash, err := cic.Hash()`?

parties/githubclient.go
```diff
@@ -149,77 +147,69 @@ func (g *GithubOp) RequestTokens(cicHash string) ([]byte, error) {
...
+	if err != nil {
+		return err
+	}
 
-	idt := pkt.OpJWSCompact()
+	idt, err := pkt.Compact(pkt.Op)
```
EthanHeilman, line 170: Should this be the package (pktoken), struct (PKToken) or the instance (pkt)? It feels like the package is best, but I can see the case for the struct. The instance makes me wonder what the particular values in the instance are influencing this output. If this is just a static method why use the instance?

parties/githubclient.go
```diff
@@ -149,77 +147,69 @@ func (g *GithubOp) RequestTokens(cicHash string) ([]byte, error) {
 	return []byte(jwt.Value), err
 }
 
-func (g *GithubOp) VerifyPKToken(pkt *pktoken.PKToken, cosPk crypto.PublicKey) (map[string]any, error) {
```
EthanHeilman, line 152: Glad you go rid of the `map[string]any` return.

EthanHeilman (changes requested):

examples/google/example.go
lgmugnier: `skidpkt` will be nil if there is an error thrown, so we check if the error was thrown.

parties/githubclient.go
lgmugnier: This is an object method because it grabs the payload from the object to put into the compact form.

parties/githubclient.go
lgmugnier: I'm good with `Hash()`, but then I would make it so that we're actually injecting meaning in the value:
```golang
	commitment, err := cic.Hash()
```

cert/ca.go
lgmugnier: All I wanted to do here is a 1:1 replacement. You had email before, so I think we should keep email again. I think the certificate logic should be revisited and this is a good point, but I was trying to keep this PR as a pure refactor.

parties/ca.go
lgmugnier: 51b721d

cert/ca.go
lgmugnier: I'm not sure I understand this, this line replaces the existing line you had with the same variable name and existing functionality. I have removed the line and am just using the map object directly, maybe that solves the issue? 299f73f

parties/githubclient.go
lgmugnier: 99d9763

parties/githubclient.go
EthanHeilman: It confused me because an instance variable is being supplied to an instance method of the same object. That said, now that I understand what you are doing I withdraw my objection. It is not how I would write it but I see the benefit of flexibility it offers.

EthanHeilman: Resolve after reading

EthanHeilman (approved):
LGTM

PR #46, opened 2023-10-26
feat: gq test modified payloads

mrjoelkamp:
* Adds tests to GQ for signature verification failure when the ID Token payload is modified
* `TestVerifyModifiedIdPayload`
  * modifies the ID Token payload before IdP signature is replaced by GQ signature
* `TestVerifyModifiedGqPayload`
  * modifies the ID Token payload after the GQ signature is generated

These tests ensure that a modified payload fails GQ signature verification in addition to the existing test of a valid payload passing GQ signature verification.

### DIFF
--- a/gq/gq_test.go
+++ b/gq/gq_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/lestrrat-go/jwx/v2/jwa"
 	"github.com/lestrrat-go/jwx/v2/jws"
 	"github.com/openpubkey/openpubkey/gq"
+	"github.com/openpubkey/openpubkey/util"
 )
 
 func TestProveVerify(t *testing.T) {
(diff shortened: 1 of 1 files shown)


gq/gq_test.go
```diff
@@ -33,10 +34,91 @@ func TestProveVerify(t *testing.T) {
...
+	}
+
+	// modify the ID Token payload to detect IdP signature invalidity via GQ verify
+	err = modifyTokenPayload(idToken, "fail")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	signerVerifier := gq.NewSignerVerifier(oidcPubKey, 256)
+	gqToken, err := signerVerifier.SignJWT(idToken)
```
jonnystoten, line 61: Just a thought for another PR - but maybe we should fail here if the `idToken` isn't valid. WDYT?

gq/gq_test.go
```diff
@@ -33,10 +34,91 @@ func TestProveVerify(t *testing.T) {
...
+		"sub": "1",
+		"iss": "test",
+		"aud": audience,
+		"iat": time.Now().Unix(),
+	}
+	modifiedPayload, err := json.Marshal(newPayload)
+	if err != nil {
+		return err
+	}
+	copy(payload, util.Base64EncodeForJWT(modifiedPayload))
```
jonnystoten, line 118: This is a bit sketchy I think - it will result in an invalid token if the old payload and new payload are not the exact same length, right? I think we should have `modifyTokenPayload` return a new payload rather than modify the existing one.

jonnystoten (commented):
Thanks, these are useful tests!

gq/gq_test.go
mrjoelkamp: agreed, refactored to create and return a new token rather than modify existing in [275ec15](https://github.com/openpubkey/openpubkey/pull/46/commits/275ec151a15c2144ac4f9e81c38c2341d2b2f9e8)

whalelines (approved):
LGTM

gq/gq_test.go
jonnystoten: Ah that's not quite what I meant, sorry - I meant maybe we should change `signerVerifier.SignJWT` to error if the idToken doesn't verify with the pubkey. Just a random musing though, not necessary for this PR. The change you made is fine too :smile:

jonnystoten (approved):
LGTM

PR #54, opened 2023-11-07
Add MFA Cosigner example

lgmugnier:
Cosigners are entities capable of additionally signing a PK token. This can happen in the same context in which the token was created, or a new one. This PR introduces a practical example for a specific type of cosigner: the MFA cosigner. 

The MFA cosigner is responsible for converting an MFA authenticator's approval into a PK token signature. A valid cosigner must have:

1. A signing key pair
2. The public key of that pair hosted as a Json Web Key Set (JWKS)

## The Cosigner's Public Key
In order to understand how issuance happens we first need to understand how the cosigner signature is verified. The cosigner has the following protected header claims:
```golang
type CosignerClaims struct {
	ID          string `json:"csid"`
	KeyID       string `json:"kid"`
	Algorithm   string `json:"alg"`
	AuthID      string `json:"eid"`
	AuthTime    int64  `json:"auth_time"`
	IssuedAt    int64  `json:"iat"` // may differ from auth_time because of refresh
	Expiration  int64  `json:"exp"`
	RedirectURI string `json:"ruri"`
}
```

The key argument we need to focus on is the `csid` or the "ID" of the Cosigner, this is what allows the verifier to verify a Cosigner's signature without being fed a public key. As stated above, it is the responsibility of every Cosigner to host their public key in the form of a JWKS. In order to tie the Cosigner's ID to its key set, the Cosigner must host their JWKS at a url with the following formula: `csid/.well-known/jwks.json`. In the example we pick a random, open port but we're hosting everything at localhost so this becomes: `http://localhost:{port}/.well-known/jwks.json`.

## Splitting it up
The MFA Cosigner is comprised of two objects: the MFA Authenticator and the Cosigner. The MFA Authenticator can be any object that abides by the following interface:
```golang
type Authenticator interface {
	Authenticate(pkt *pktoken.PKToken) error
	URI() string
}
```
The result of `Authenticate()` is a binary, either the user is authenticated by MFA or they are not (`err != nil`).

## WebAuthn
This specific MFA example uses WebAuthn as its MFA mechanism and because of the nature of this example in its current form, it forces the user to both register and login consecutively. This might feel dumb but it is necessary for the proper functioning of WebAuthn, I just don't recommend using it as is for any kind of practical application.


EthanHeilman:
This PR looks great. I wanted to discuss an architecture issue outside the code review since I think it is important enough to  document our thoughts and discussion on the package structure.

 When i first read this I assumed that `client.cosigner` was the code the client used to talk to the cosigner similar to `client.provider` being the code the client uses to talk to the OpenID Providers. However this code creates cos token (i.e. cosigner claims) and signs it as the cosigner.

I think about OpenPubkey as having the following parties:
* **The client:**
   *  creates CIC 
   *  requests ID Tokens from OpenID Providers 
   *  IF desired: requests signatures from cosigners to get  Cosigner Claims (COS)
   * Uses ID, CIC and COS to build PK Tokens
   * IF desired: Runs refresh flows
   * Signs messages under a PK Token
* **The verifier/PK Token:**
   * Receives PK Token's and signed messages,
   * Uses OpenID Providers and Cosigners public keys to verify signed messages under a PK Token
* **The provider (OpenID Provider):**
   * Authenticates identities issues ID Tokens, Refresh Tokens
 * **The Cosigner:** 
   * Authenticates identities issues cosigner claims (COS Tokens)

I would propose the cosigner is moved outside the client package either:

(a) following the example of the PK Token which contains the verifier party and having the cosigner live at the root of the module:
`github.com/openpubkey/openpubkey/client/cosigner/` --> github.com/openpubkey/openpubkey/cosigner/

(b). or putting both the client and the cosigner in a package called parties: `github.com/openpubkey/openpubkey/parties/client/` 
`github.com/openpubkey/openpubkey/parties/cosigner/mfa`

I personally favor (a) over (b) because (a) seems the least disruptive to existing file locations/package structure and I like shorter package paths.

EthanHeilman:
What if we renamed `cosid` to `iss` and then followed the OpenID Connect discovery spec that says:
```
OpenID Providers supporting Discovery MUST make a JSON document available at the path formed by concatenating the string /.well-known/openid-configuration to the Issuer
``` https://openid.net/specs/openid-connect-discovery-1_0.html

This fits really well with work you did in the PR having the cosigner have a JWKS URI and using the `/.well-known` scheme (RFC-5785)

I choose the name `cosid` when I knew much less about OIDC than I do now. I believe you or someone else at a point in the past suggested this change a while ago, but I was worried about OPs and COS impersonating each other. I know realize that isn't a problem because if a malicious put an `iss=https://accounts.google.com` in a protected header, the validator would look up Google's public key at Google's JWKS URI and see that Google didn't sign it.

[RFC-5785: Defining Well-Known Uniform Resource Identifiers (URIs)](https://datatracker.ietf.org/doc/html/rfc5785)

File: examples/mfa/webauthn/server.go [hunk 1]
```diff
@@ -0,0 +1,204 @@
...
+	}()
+
+	return server, nil
+}
+
+func (s *Server) URI() string {
+	return s.uri
+}
+
+func (s *Server) Authenticate(pkt *pktoken.PKToken) error {
```
EthanHeilman, line 79: IIRC we discussed that since we are still thinking through OpenPubkey POP Auth, that the example would not perform OpenPubkey authentication but only MFA authentication. I'm good with this.

To ensure we don't lose track of this, can you create a comment in the code and a github issue explaining that the MFA Auth example currently just naively trusts the identity in the PK Token. Then when we add OpenPubkey POP Auth to the MFA cosigner we can remove that comment and close that issue.

examples/mfa/webauthn/server.go
```diff
@@ -0,0 +1,204 @@
...
+	server.auth, err = webauthn.New(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	http.Handle("/", http.FileServer(http.Dir("webauthn/static")))
+
+	http.HandleFunc("/check-registration", server.checkIfRegistered)
+
+	http.HandleFunc("/register/begin", server.beginRegistration)
```
EthanHeilman, line 60: This is a non-blocker for this PR and example, but I created an issue to think about how we prevent CSRF or if this is safe. 
https://github.com/openpubkey/openpubkey/issues/57

Resolve after reading

examples/mfa/webauthn/static/index.html
```diff
@@ -0,0 +1,165 @@
...
+                method: 'GET',
+            });
+
+            let data = response.json();
+
+            if (!data.isRegistered) {
+                try {
+                    await register();
+                } catch(error) {
+                    statusDiv.textContent = `Error during registration: ${error}`;
```
EthanHeilman, line 134: Any danger of XSS attack in error? Is error guarantied to be escaped?

EthanHeilman (changes requested):
Ready to merge after:
* Addressing my comment on package location of the cosigner.
* Determining if error XSS is a real concern or not.
* Creating issue and comment explaining MFA Cosigner is only doing MFA auth.

EthanHeilman:
I tested this on Windows and OSX using the brave browser.

On OSX I tested this with a USB-C yubikey and the macbook fingerprint reader functioning as a passkey. Both worked.

On windows I couldn't get any MFA device to work. The webauthn screen would appear but it would never recognize that a webauthn device was plugged in. I suspect that this has nothing to do with your code but rather that windows really didn't seem to like the fact that the cosigner was running on localhost. I'll investigate further but I don't think this should block the PR.

examples/mfa/webauthn/static/index.html
lgmugnier: I hear what you're saying about XSRF and XSS, but I don't think those should be a concern in an example which isn't meant for practical application and whose preventative details wouldn't have an impact on opk

examples/mfa/webauthn/server.go
lgmugnier: cd30ca8

jonnystoten:
Worked for me with Firefox on Linux :tada:

> This specific MFA example uses WebAuthn as its MFA mechanism and because of the nature of this example in its current form, it forces the user to both register and login consecutively

This was a bit confusing, not a blocker but maybe we could display something on the page to indicate that this is happening?

cosigner/mfa/cosigner.go
```diff
@@ -0,0 +1,66 @@
...
+func (c *Cosigner) Cosign(pkt *pktoken.PKToken) error {
+	if err := c.mfa.Authenticate(pkt); err != nil {
+		return err
+	}
+
+	protected := pktoken.CosignerClaims{
+		ID:          c.csid,
+		KeyID:       c.kid,
+		Algorithm:   c.alg.String(),
+		AuthID:      "12345678",
```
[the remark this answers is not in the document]
EthanHeilman: AuthID in the OpenPubkey paper is an auth session identifier so that if the MFA Cosigner logged details about when it granted the signature, the signature/cosigner claims in a PK Token could be linked back to a specific time and place.

This enables the following features:

* A user could request that the Cosigner revoke a particular auth session/PK Token, much like users in gmail can revoke particular gmail sessions.

* A forensic investigator or incident response, could track a compromise and say see what MFA device was used to authenticate to the MFA Cosigner and determine which MFA devices were compromised.

* Risk and Anomaly based revocation of particular PK Tokens associated with particular sessions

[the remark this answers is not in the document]
lgmugnier: @jonnystoten you do ask a good question. I would like to revisit all the claims at some point in the future when the architecture and POP Auth have been implemented so we can better understand what's necessary. For now, I implemented them verbatim from the paper.

examples/mfa/webauthn/static/index.html
EthanHeilman: We should avoid XSS in example javascript in case someone uses them and to establish patterns of security. If something is unsafe in an example we should leave a comment.

I did some digging and `.textContent` naturally protects against XSS and can safely be used.

> RULE #6 - Populate the DOM using safe JavaScript functions or properties The most fundamental safe way to populate the DOM with untrusted data is to use the safe assignment property textContent.

[OWASP: DOM based XSS Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.html#rule-6-populate-the-dom-using-safe-javascript-functions-or-properties)

I consider my comment here addressed and I am convinced this is secure.

EthanHeilman (approved):
You have addressed all my comments, looks good to merge

cosigner/mfa/cosigner.go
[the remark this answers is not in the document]
EthanHeilman: If you don't mind @lgmugnier I might address this issue in a PR that builds on this one, after this is merged.

jonnystoten (approved):
LGTM :rocket:

Issue #62, opened 2023-11-16
Error using google login example


jonnystoten:
As mentioned by @lgmugnier on the community call yesterday, there's an issue with the Google login example. The login process in the browser seems to work correctly but we get an error on the library side:

```
❯ go run example.go login
INFO[0000] listening on http://localhost:3000/          
INFO[0000] press ctrl+c to stop                         
Error logging in: error verifying PK Token: failed to verify signature from OIDC provider: error verifying OP signature on PK Token (ID Token invalid): invalid signature (signature verification failed: square/go-jose: error in cryptographic primitive)
```

lgmugnier:
I'm looking into this today

lgmugnier:
Fix here: https://github.com/openpubkey/openpubkey/pull/63

PR #63, opened 2023-11-16
Allow setting op signature type on pk token creation

lgmugnier:
Before, we were assuming that the PK token was always being created with the oidp's ID token. That assumption changed in #48, but the code was still setting the signature type as `oidc` instead of `oidc_gq`. Therefore, the validator thought the provider's signature was non-gq and went through that flow where it failed because the signature was gq.

Relates to issue https://github.com/openpubkey/openpubkey/issues/62

### DIFF
--- a/pktoken/pktoken.go
+++ b/pktoken/pktoken.go
@@ -36,9 +36,15 @@ type PKToken struct {
 	Cos     *Signature // Cosigner Signature
 }
 
-func New(idToken []byte, cicToken []byte) (*PKToken, error) {
+func New(idToken []byte, cicToken []byte, isGQ bool) (*PKToken, error) {
 	pkt := &PKToken{}
-	if err := pkt.AddSignature(idToken, Oidc); err != nil {
+
+	sigType := Oidc
+	if isGQ {
+		sigType = Gq
+	}
+
+	if err := pkt.AddSignature(idToken, sigType); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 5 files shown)


pktoken/pktoken.go
```diff
@@ -36,9 +36,15 @@ type PKToken struct {
 	Cos     *Signature // Cosigner Signature
 }
 
-func New(idToken []byte, cicToken []byte) (*PKToken, error) {
+func New(idToken []byte, cicToken []byte, isGQ bool) (*PKToken, error) {
```
EthanHeilman, line 39: What do you think about?
```golang
func New(idToken []byte, cicToken []byte, sigType SignatureType,) 
```

I like the property that this would document what is being flagged
``` 
pktoken.New(idToken.Bytes(), cicToken, pktoken.Gq)
```
vs a naked Boolean
``` 
pktoken.New(idToken.Bytes(), cicToken, false)
```

Just a idea, not a blocking request

EthanHeilman:
Thanks for taking a look at this and figuring how the bug was introduced. How tricky would it be to unit test this so that we catch regressions?

lgmugnier:
> Thanks for taking a look at this and figuring how the bug was introduced. How tricky would it be to unit test this so that we catch regressions?

Found a super easy way to do this. I just took our existing tests and had them check the provider's signature type instead of assuming it based on test data. d4fabf2

pktoken/pktoken.go
lgmugnier: I have some small issues with the signature type and exposing it like that. Ideally, I would like to be able to rely on the `arg` value in the header to figure out whether or not it's a GQ signature and verify based on that instead of an unprotected value. So, for now I want to keep it a bool because that's what we have elsewhere and hopefully we won't need that in the future.

EthanHeilman (approved):

PR #66, opened 2023-11-21
Use GQ256 alg for GQ signatures

### LINKED ISSUE #15: GQ Signature flagging

jonnystoten:
Closes #15.

For GQ signatures, we take the original ID token from the OIDC provider and replace the header. The new header is the same as the original with two changes:

1. The `alg` claim is set to `GQ256`
2. The `kid` claim is set to the original header value from the ID token. This value is taken directly from the JWT token signing payload (raw base64 encoded)

The `alg` value gives us an easy way to identify the GQ signature from a JWS containing multiple signatures. The `kid` claim allows us to recreate the original signing payload in order to verify the signature.

For example, an ID token header of
```json
{"alg":"RS256","typ":"JWT"}
```
would be replaced with a new header:
```json
{"alg":"GQ256","kid":"eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCJ9","typ":"JWT"}
```

### DIFF
--- a/client/oidc.go
+++ b/client/oidc.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/awnumar/memguard"
 	"github.com/lestrrat-go/jwx/v2/jws"
+	"github.com/openpubkey/openpubkey/gq"
 	"github.com/openpubkey/openpubkey/pktoken"
 	"github.com/openpubkey/openpubkey/util"
 )
@@ -93,8 +94,15 @@ func VerifyPKToken(ctx context.Context, pkt *pktoken.PKToken, provider OpenIdPro
 
 	switch sigType {
 	case pktoken.Gq:
+		origHeaders, err := gq.OriginalJWTHeaders(idt)
+		if err != nil {
+			return err
+		}
+
+		fmt.Println(string(origHeaders))
+
 		// TODO: this needs to get the public key from a log of historic public keys based on the iat time in the token
-		pubKey, err := provider.PublicKey(ctx, idt)
+		pubKey, err := provider.PublicKey(ctx, origHeaders)
 		if err != nil {
 			return fmt.Errorf("failed to get OP public key: %w", err)
 		}
--- a/client/oidc.go
+++ b/client/oidc.go
@@ -89,7 +89,16 @@ func VerifyPKToken(ctx context.Context, pkt *pktoken.PKToken, provider OpenIdPro
 
 	idt, err := pkt.Compact(pkt.Op)
 	if err != nil {
-		return fmt.Errorf("")
+		return err
+	}
+
+	issuer, err := ExtractClaim(idt, "iss")
+	if err != nil {
+		return err
+	}
+
+	if issuer != provider.Issuer() {
+		return fmt.Errorf("expected token to have issuer %s, got %s", provider.Issuer(), issuer)
 	}
 
 	alg, ok := pkt.ProviderAlgorithm()
--- a/client/oidc.go
+++ b/client/oidc.go
@@ -71,7 +71,7 @@ func (id *OidcClaims) UnmarshalJSON(data []byte) error {
 type OpenIdProvider interface {
 	Issuer() string
 	RequestTokens(ctx context.Context, cicHash string) (*memguard.LockedBuffer, error)
-	PublicKey(ctx context.Context, headers map[string]any) (crypto.PublicKey, error)
+	PublicKey(ctx context.Context, headers jws.Headers) (crypto.PublicKey, error)
 	VerifyCICHash(ctx context.Context, idt []byte, expectedCICHash string) error
 	VerifyNonGQSig(ctx context.Context, idt []byte, expectedNonce string) error
 }
@@ -108,32 +108,24 @@ func VerifyPKToken(ctx context.Context, pkt *pktoken.PKToken, provider OpenIdPro
 
 	switch alg {
 	case gq.GQ256:
-		origHeaders, err := gq.OriginalJWTHeaders(idt)
+		origHeadersB64, err := gq.OriginalJWTHeaders(idt)
 		if err != nil {
 			return err
 		}
 
-		origHeaderClaims, err := parseJWTSegment(origHeaders)
+		origHeaders := jws.NewHeaders()
+		err = parseJWTSegment(origHeadersB64, &origHeaders)
 		if err != nil {
 			return err
 		}
 
-		algClaim, ok := origHeaderClaims["alg"]
-		if !ok {
-			return fmt.Errorf("missing alg claim")
-		}
-
-		alg, ok := algClaim.(string)
-		if !ok {
-			return fmt.Errorf("expected alg claim to contain a SignatureAlgorithm, got %T", algClaim)
-		}
-
-		if jwa.SignatureAlgorithm(alg) != jwa.RS256 {
+		alg := origHeaders.Algorithm()
+		if alg != jwa.RS256 {
 			return fmt.Errorf("expected original headers to contain RS256 alg, got %s", alg)
 		}
 
 		// TODO: this needs to get the public key from a log of historic public keys based on the iat time in the token
-		pubKey, err := provider.PublicKey(ctx, origHeaderClaims)
+		pubKey, err := provider.PublicKey(ctx, origHeaders)
 		if err != nil {
 			return fmt.Errorf("failed to get OP public key: %w", err)
 		}
@@ -173,17 +165,7 @@ func VerifyPKToken(ctx context.Context, pkt *pktoken.PKToken, provider OpenIdPro
 	return nil
 }
 
-func DiscoverPublicKey(ctx context.Context, headers map[string]any, issuer string) (crypto.PublicKey, error) {
-	kidRaw, ok := headers["kid"]
-	if !ok {
-		return nil, fmt.Errorf("missing kid claim")
-	}
-
-	kid, ok := kidRaw.(string)
-	if !ok {
-		return nil, fmt.Errorf("expected kid claim to be a string, got %T", kidRaw)
-	}
-
+func DiscoverPublicKey(ctx context.Context, headers jws.Headers, issuer string) (crypto.PublicKey, error) {
 	discConf, err := oidcclient.Discover(issuer, http.DefaultClient)
 	if err != nil {
 		return nil, fmt.Errorf("failed to call OIDC discovery endpoint: %w", err)
@@ -194,6 +176,7 @@ func DiscoverPublicKey(ctx context.Context, headers map[string]any, issuer strin
 		return nil, fmt.Errorf("failed to fetch to JWKS: %w", err)
 	}
 
+	kid := headers.KeyID()
 	key, ok := jwks.LookupKeyID(kid)
 	if !ok {
 		return nil, fmt.Errorf("key %q isn't in JWKS", kid)
@@ -218,7 +201,8 @@ func ExtractClaim(idt []byte, claimName string) (string, error) {
 		return "", fmt.Errorf("failed to split/decode JWT: %w", err)
 	}
 
-	payload, err := parseJWTSegment(payloadB64)
+	payload := make(map[string]any)
+	err = parseJWTSegment(payloadB64, &payload)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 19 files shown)


client/providers/google.go
```diff
@@ -120,12 +120,19 @@ func (g *GoogleOp) VerifyCICHash(ctx context.Context, idt []byte, expectedCICHas
...
-func (g *GoogleOp) PublicKey(ctx context.Context, idt []byte) (crypto.PublicKey, error) {
-	j, err := jws.Parse(idt)
+func (g *GoogleOp) PublicKey(ctx context.Context, headersEnc []byte) (crypto.PublicKey, error) {
+	headersJSON, err := util.Base64DecodeForJWT(headersEnc)
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse JWS: %w", err)
+		return nil, err
+	}
+
+	headers := make(map[string]any)
```
whalelines, line 129: Why would the headers not be `map[string]string`? Is something unmarshalling the strings into something else?

whalelines, line 129: Why would the headers not be `map[string]string`? Is something unmarshaling the strings into something else?

whalelines (commented):
Is part of the un-drafting to add tests?

client/providers/google.go
jonnystoten: You're right, this is just me being lazy

jonnystoten:
> Is part of the un-drafting to add tests?

Yes!

client/providers/google.go
whalelines: I don't know if forcing yourself to cast to string every time you need to use it counts as laziness.

EthanHeilman:
Instead of using the JOSE header `orig` to specify the original protected header, what about storing the original protected header in the `kid`.

```golang
headers["orig"] = string(origHeaders)
```
vs.

```golang
headers["kid"] = string(origHeaders)
```

**Pros:**
1. Avoids definining a novel JOSE header, `orig`, Some libraries that only support known JOSE headers.
2. `kid` is designed to specify a JWK which is what what the original protected header itself does. It is common to assign `kid` as a Public key hashes and fingerprints

**Cons:**
1. Could be confusing as it overloads the meaning of the `kid`. The header name `orig` has a more explicit meaning. Generally you don't deserialize a `kid`.

>The "kid" (key ID) Header Parameter is a hint indicating which key was used to secure the JWS.  This parameter allows originators to explicitly signal a change of key to recipients.
https://www.rfc-editor.org/rfc/rfc7515.html

jonnystoten:
I think that's a good idea. It will make it slightly easier to use the jwx library to get the original headers out - the library makes it hard to use non-standard headers.

You trimmed your quote early there, the next sentence is

> The structure of the "kid" value is unspecified.

which means it would be valid according to the spec.

client/providers/google.go
jonnystoten: I originally changed this to unmarshal to a struct, but on second thought, I think the map[string]any is better. This is untrusted input so it really could be anything, and when unmarshalling into a struct, if the key is missing we'll just get an empty string. With a map we can check it's there and the correct type.

client/oidc.go
```diff
@@ -89,6 +99,52 @@ func VerifyPKToken(ctx context.Context, pkt *pktoken.PKToken, provider OpenIdPro
 	return nil
 }
 
+func DiscoverPublicKey(ctx context.Context, headersEnc []byte, issuer string) (crypto.PublicKey, error) {
```
EthanHeilman, line 102: I've tried to use a postfix of `B64` to signal that a variable is Base64 encoded and I'd like to maintain that as a convention: `headersEnc` to `headersB64`.

client/oidc.go
```diff
@@ -89,6 +99,52 @@ func VerifyPKToken(ctx context.Context, pkt *pktoken.PKToken, provider OpenIdPro
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch to JWKS: %w", err)
+	}
+
+	key, ok := jwks.LookupKeyID(kid)
+	if !ok {
+		return nil, fmt.Errorf("key %q isn't in JWKS", kid)
+	}
+
+	pubKey := new(rsa.PublicKey)
```
EthanHeilman, line 139: Before casting to a particular key type, say RSA, we should check that the algorithm of the returned JWK matches the key type we are casting to. This is similar to my comment above, but unless I am mistaken, I think we should do this check in both places.

Your PR does not introduce this issue as it exists in the current code, but it is a valuable check to have.

pktoken/pktoken.go
```diff
@@ -192,31 +198,39 @@ func (p *PKToken) UnmarshalJSON(data []byte) error {
...
+		if len(pubMap) == 0 {
+			signature.SetPublicHeaders(nil)
 		}
 
-		sigHeaderString, ok := sigHeader.(string)
+		protected := signature.ProtectedHeaders()
+		typeHeader, ok := protected.Get(jws.TypeKey)
+		if !ok {
+			// TODO: default to OIDC?
+			return fmt.Errorf("missing typ header")
```
EthanHeilman, line 213: I believe should default to OIDC here since it is completely within the JWS standard to not include a `typ` header. OpenPubkey should always include a `typ` header for a COS and CIC signature.

It might make sense to log a warning, because it is very uncommon for a OP to not include a `typ` header. It is happens, it is probably a since of some weirdnesss.

EthanHeilman (changes requested):
Not a full complete review, but I wanted to write these two issues down

client/oidc.go
jonnystoten: I think this is already OK because the jwx library does this check itself. The key is parsed after being fetched and stored in an appropriate struct based on the `alg` field (see https://github.com/lestrrat-go/jwx/blob/262da1ab53fc42371eb8a13ae61044589b1a239b/jwk/jwk.go#L443-L467). Then `key.Raw(pubKey)` checks that `pubKey` is of the appropriate type for the type of `key` and errors otherwise.

client/oidc.go
```diff
@@ -41,15 +46,20 @@ func VerifyPKToken(ctx context.Context, pkt *pktoken.PKToken, provider OpenIdPro
...
+	switch alg {
+	case gq.GQ256:
+		origHeaders, err := gq.OriginalJWTHeaders(idt)
+		if err != nil {
+			return err
+		}
+
 		// TODO: this needs to get the public key from a log of historic public keys based on the iat time in the token
-		pubKey, err := provider.PublicKey(ctx, idt)
+		pubKey, err := provider.PublicKey(ctx, origHeaders)
```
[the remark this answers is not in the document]
jonnystoten: For (1) I think we should first check that the `alg` claim in `origHeaders` is RS256, then just check that the key returned from `provider.PublicKey` is actually `*rsa.PublicKey`. At the moment this code will panic if it isn't!

For (2), totally agree, I'll add that check.

client/oidc.go
```diff
@@ -89,6 +99,52 @@ func VerifyPKToken(ctx context.Context, pkt *pktoken.PKToken, provider OpenIdPro
...
+	jwks, err := jwk.Fetch(ctx, discConf.JwksURI)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch to JWKS: %w", err)
+	}
+
+	key, ok := jwks.LookupKeyID(kid)
+	if !ok {
+		return nil, fmt.Errorf("key %q isn't in JWKS", kid)
+	}
+
```
EthanHeilman, line 138:
```suggestion
	if key.Algorithm() != jwa.SignatureAlgorithm.RS256 {
		return nil, fmt.Errorf("unsupported alg (%q) in JWK with kid (%q) for OP (%q), OpenPubkey only alg=RS256 supported: \n", key.Algorithm(), kid, issuer)
	}
```

client/oidc.go
EthanHeilman: You are correct I misread what the code was doing here.

It might save a future dev some time if we create a custom error for the case where the key is not RS256. I've made a suggestion for this above. Not a requirement for this PR so feel free to reject the suggestion and resolve this comment.

jonnystoten: Yes, good idea. I'll tweak the suggestion slightly

EthanHeilman:
Manually tested this using the google example with GQ signatures on and off and inspected the PK Tokens. Every thing works and looks as expected.

Details included only for the sake of posterity:

OP signature pheader (GQ=true)
```json
{"alg":"GQ256","kid":"eyJhbGciOiJSUzI1NiIsImtpZCI6IjliMDI4NWMzMWJmZDhiMDQwZTAzMTU3YjE5YzRlOTYwYmRjMTBjNmYiLCJ0eXAiOiJKV1QifQ","typ":"JWT"}
```

Decoded signature from `kid`
```json
{"alg":"RS256","kid":"9b0285c31bfd8b040e03157b19c4e960bdc10c6f","typ":"JWT"}
```

CIC signature:
```json
{"alg":"ES256","extra":"yes","rz":"ac8dbe1734c088b118b6a7e9fd0a6d9f1142ee6ff0d5b9eea9d2fac10299e135","typ":"CIC","upk":{"alg":"ES256","crv":"P-256","kty":"EC","x":"ApbgXJU728FDasOsnZs2uVxzmLT6UosyQoYCunuBhTk","y":"B5b5U9C_kzOr5ta06-ffZnZDYK6OoJXT9WW3cQjbYrg"}}
```
We do include the alg in both upk JWK object and at the pheader which is annoying. Might be worth addressing that at some point, but not in this PR.

OP signature pheader (GQ=false)
```json
{"alg":"RS256","kid":"9b0285c31bfd8b040e03157b19c4e960bdc10c6f","typ":"JWT"}
```

EthanHeilman (approved):
Tested manually, LGTM

EthanHeilman (approved):
LGTM

PR #70, opened 2023-12-06
Run unit tests against ISO/IEC 14888-2:2008 GQ test vectors

### LINKED ISSUE #69: Run unit tests against ISO/IEC 14888-2:2008 GQ test vectors
As discussed in #67, we can further validate our GQ implementation by adding test cases against the specific numerical values in the ISO standard. 

This will require some refactoring of the existing code because the ISO tests use SHA-1/PSS as a format mechanism, whereas we use SHAKE-256/PKCSv15.

johncmerfeld:
Closes #69 

Adds specific signer/verifier tests against the values laid out in the ISO standard on which our implementation is based.

There is one painful difference between the standard and our implementation. It uses SHA-1 instead of SHAKE-256 as a hash function and uses PSS instead of PKCS15 as an encoding / format mechanism. I considered making our solution fully plug-and-play with regard to these functions. However, this incurs a huge code complexity expense because the functions in question do not share the same interface in the go standard library. Instead, I opted to turn the relevant functions into `vars` and rewire them at test time. This requires moving the tests from a dedicated `gq_test` package to the `gq` package.

The above is all somewhat messy and I am open to discussing alternatives. However, I believe it is the cleanest way of achieving this test coverage. The good news is that **our implementation was correct on the first attempt with no modifications needed**.

### DIFF
--- a/gq/gq_iso_test.go
+++ b/gq/gq_iso_test.go
@@ -29,8 +29,6 @@ var svISO signerVerifier
 
 // Test our signer using the values specified in ISO/IEC 14888-2:2008
 func TestSignerISO(t *testing.T) {
-	initializeISOValues(t)
-
 	// The test vector specifies that h(W||M) use SHA-1
 	h := hash
 	hash = sha1Hash
@@ -62,8 +60,6 @@ func TestSignerISO(t *testing.T) {
 
 // Test our verifier using the values specified in ISO/IEC 14888-2:2008
 func TestVerifierISO(t *testing.T) {
-	initializeISOValues(t)
-
 	// The test vector specifies that h(W||M) use SHA-1
 	h := hash
 	hash = sha1Hash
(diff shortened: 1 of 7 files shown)


EthanHeilman (approved):

gq/gq_iso_test.go
```diff
@@ -0,0 +1,162 @@
...
+
+	encodedSigISO := util.Base64EncodeForJWT(sigISO)
+	ok := svISO.Verify(encodedSigISO, idISO, mISO)
+
+	if !ok {
+		t.Fatal("signature verification failed")
+	}
+}
+
+func initializeISOValues(t *testing.T) {
```
jonnystoten, line 90: This is minor and non-blocking, but I think it might be nice to change this to an init function:

```suggestion
func init() {
```

That way it will be called automatically once rather than us having to do it manually in each test. The disadvantage is that it won't have access to the `testing.T` so we'll have to panic on any errors. WDYT?

gq/gq_iso_test.go
```diff
@@ -0,0 +1,162 @@
...
+	nBytes, err := hexToBytes(nHex)
+	if err != nil {
+		t.Fatal(err)
+	}
+	n, err := bigmod.NewModulusFromBig(new(big.Int).SetBytes(nBytes))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	svISO = signerVerifier{n, v, 128, 10, 1}
```
jonnystoten, line 106: Could you please add labels to the fields here so it's easier to see which parameter is which?

gq/gq_iso_test.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	r, err := bigmod.NewNat().SetBytes(rRaw, n)
+	if err != nil {
+		return nil, err
+	}
+
+	ys[0] = r
+	return ys, nil
+}
```
jonnystoten, line 144: Not sure I fully understand this part. `ys` is initialized to have length `t` but we only fill out the first one, so the rest are nil. Should the length just be 1?

jonnystoten (changes requested):
This is awesome, thanks!

Just a few minor comments below

gq/gq_iso_test.go
whalelines: You can use `TestMain` so as not to pollute the `init` space, https://pkg.go.dev/testing#hdr-Main

johncmerfeld: I was on the fence about this and you've convinced me that `init()` is better -- it's definitely more idiomatic and `t.Fatal` isn't really any better than `panic` so I'm good with it!

gq/gq_iso_test.go
johncmerfeld: Good callout, thanks!

jonnystoten (approved):
LGTM, thanks!

PR #71, opened 2023-12-06
Add support for external Op public key resolution

kipz:
To be able to verify signatures once Op public keys have rotated, it's necessary for the verifier to consult a log.

This PR adds a wrapper/decorator so that public key resolution can be delegated to a custom function.

It might make more sense to make breaking changes to the provider APIs rather than wrapping like this given we are not at 1.0 yet though.

### DIFF
--- a/client/providers/public.go
+++ b/client/providers/public.go
@@ -1,0 +1,30 @@
+package providers
+
+import (
+	"context"
+	"crypto"
+	"fmt"
+
+	"github.com/openpubkey/openpubkey/client"
+)
+
+type PublicKeyResolver func(context.Context, []byte) (crypto.PublicKey, error)
+
+type publicKeyResolver struct {
+	client.OpenIdProvider
+	resolver PublicKeyResolver
+}
+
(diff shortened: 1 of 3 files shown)


EthanHeilman:
I'm good with making breaking changes to provider APIs.

client/providers/public.go
```diff
@@ -0,0 +1,30 @@
...
+
+	"github.com/openpubkey/openpubkey/client"
+)
+
+type PublicKeyResolver func(context.Context, []byte) (crypto.PublicKey, error)
+
+type publicKeyResolver struct {
+	client.OpenIdProvider
+	resolver PublicKeyResolver
+}
```
jonnystoten, line 16: I'm not sure this adds much - I think it's reasonable for the calling code to define their own struct which embeds `client.OpenIdProvider` and add their own implementation of `PublicKey`. WDYT?

jonnystoten:
Yeah I agree, breaking changes are OK at this stage. What did you have in mind?

kipz:
Thanks both. Closing for now as clients can just as easily decorate

Issue #34, opened 2023-10-11
PK Token Properties


lgmugnier:
I want to have a discussion about what our goals are with the PK Token because it impacts data structures which are always nice to get as right as possible, as soon as possible. 

I propose that the PK Token should have the following three properties:

**Property 1**: The PK Token does not require any outside information in order to verify itself
- This isn't to say that external code shouldn't verify certain claims or parameters. For example, it would still be best practices to check that the issuer claim or a key algorithm match expected values, but a PK Token would only use its own values for verification.

**Property 2**: A valid PK Token should have, at a minimum:
- A single oidc provider signature whether that be in the form of rsa or gq (or other!)
- A single user signature over the same payload including client instance claims headers

**Property 3**: The PK Token supports any number of cosigners
- A PK Token can have zero or more cosigners without violating **Property 1**

jonnystoten:
Apologies for the late comment!

**Property 1**: I _think_ this is fine but this stuff is tricky. I was worried that not requiring alg be checked against an externally provided value could be dangerous, but I think we're OK because we require use of JWK, so the alg in the JWK must match the alg in the JWS. Does that make sense?

**Property 2**: You mean we require one OP signature but we allow more than one and we require one user signature but we allow more than one, right? It would be strange to have multiple OP signatures I think, is there a use case for that? Multiple user signatures makes sense if you want to bind multiple keys.

**Property 3**: Makes sense to me!

lgmugnier:
> Property 1: I think this is fine but this stuff is tricky. I was worried that not requiring alg be checked against an externally provided value could be dangerous, but I think we're OK because we require use of JWK, so the alg in the JWK must match the alg in the JWS. Does that make sense?

The emphasis I want to make with this property is the division of responsibilities. To your point, the internal pktoken validation might check that the JWK's algorithm is the same as the `alg` claim, that makes it "internally valid". But there is this second set of responsibilities like doing that external check on whether the value of the `alg` claim is acceptable. For example, if `alg` was "None" that's technically an accepted value so the pktoken would still _internally_ verify, but that would be bad if that were the case so it may not be _externally_ valid.

What I really want to accomplish with this property is to force this idea that you can validate the pk token anywhere, for the most part, like other json web things. Sure, you might not check whether it was issued by "google" instead of "evil.co" or whether the signing algorithm is "None", but if a pk token lands in your lap, you can validate it.

[...]

PR #72, opened 2023-12-07
Creates PK Token validation object for allow listing OPs and Cosigners

EthanHeilman:
When validating a PK Token currently we just check that OP Issuer and OP Audience in the ID Token matches the issuer and audience in the Provider struct and the OP and Cosigner signature validates. The current approach has three limitations:

1. Only one OP can be provided to `client.VerifyPKToken(cxt, pkt, OP)`. This means if a verifying party wanted to accept two or more OPs, `op1, op2, ... opN1`, they would have call: `client.VerifyPKToken(cxt, pkt, op1)`, ... `client.VerifyPKToken(cxt, pkt, opN)`. Each of these would involve downloading each of the OP public keys and doing signature validation.

2. More of an issue is that `client.VerifyPKToken` does not provide a parameter to supply the allowed cosigners. As such any cosigner will be accepted.

3. If PK Token does not include a Cosigner signature, then it just automatically bypasses the cosigner verification. This is because the existing code does not have a way to signal that a cosigner is or is not required.

In this PR we solve both of these problems by creating a PKTokenVerifer struct which allows specifying an allowlist of OPs and an allowlist of cosigners. This allowlist is checked against the PK Token before any remote JWKS calls are made to OPs or Cosigners unless the PK Token uses an OP issuer and Cosigner issuer that is on the allowlist.

```golang
// Verify our pktoken including the cosigner signature
		err = client.PKTokenVerifer{
			AllowedProviders: []client.OpenIdProvider{provider},
			AllowedCosigners: []client.CosignerProvider{cosignerProvider},
		}.Verify(cxt, pkt)
```

If a Cosigner requirement is not desired then that is signaled by not setting the AllowedCosigners list.

```golang
// Verify our pktoken including the cosigner signature
		err = client.PKTokenVerifer{
			AllowedProviders: []client.OpenIdProvider{provider},
		}.Verify(cxt, pkt)
```

## Open Questions:

1. What is the right place for this verifier to live. Currently we have `VerifyPKToken` live in client, but that does not seem completely right as client assumes the ability to sign. Should live in pktoken?

2. What is the best way to flag cosigner required or not required during verification? Are there existing golang OIDC patterns we could use?

3. Should this verifier be the place where we add [external OP public key resolution](https://github.com/openpubkey/openpubkey/pull/71) for things like pulling an OP public key from a TUF or transparency log? If so, what changes do we need to make to verifier to support this?

## Notes

This draft PR has an lots of code because it depends on #68 . I broke it from that PR because it felt separate and I wanted more time to think about the right place to put this code. It should only be reviewed after https://github.com/openpubkey/openpubkey/pull/68 is merged as it depends on it.

The only new code in this PR lives here:

https://github.com/openpubkey/openpubkey/pull/72/files#diff-7c00b0071153c0d4539ede345ce0628b6d9b4f83ac2e44cb291284a6c2571dcbR47-R145


lgmugnier:
Based on the questions you have here, I think it would be pertinent to figure out the solution to [this](https://github.com/openpubkey/openpubkey/issues/34) issue first which touches a lot on the different aspects of verification and who's responsible for what. For example, the concept of "AllowedCosigners" and "AllowedProviders" should probably live in the client since it's already configured with those whereas the pk token will verify based on whatever cosigners and providers it has.

EthanHeilman:
The verifier struct has been moved in #68.
Thus, I am closing this PR.

PR #79, opened 2023-12-21
refactor: ca and cert packages

mrjoelkamp:
## Summary
* Combines the original `ca` and `cert` packages into one `ca` package with tests
* Uses more generic `CreateX509Cert(pkToken *pktoken.PKToken, signer crypto.Signer) ([]byte, error)` for the `cert` package
  * taken from https://github.com/openpubkey/signed-attestation/pull/4

This PR doesn't meet any goals for the smuggler interface but moves us in the right direction. The main motivation for this change is to move this common reusable `CreateX509Cert()` from the `signed-attestation` repo into `openpubkey`


ca/ca.go
```diff
@@ -26,7 +26,7 @@ import (
 
 // TODO: make requiredAudience a configuration option
 var (
-	requiredAudience = "184968138938-g1fddl5tglo7mnlbdak8hbsqhhf79f32.apps.googleusercontent.com"
+	requiredAudience = "also_me"
```
whalelines, line 29: What's going on here?

ca/ca_test.go
```diff
@@ -70,19 +78,30 @@ func TestCertCreation(t *testing.T) {
...
 
-	readCert, err := util.ReadCertFile(certPath)
+	if string(ca.CaCertBytes) != string(testCa.CaCertBytes) {
+		t.Fatal("failed reading CA cert bytes from disk")
+	}
+
+	ecPub, err := x509.MarshalPKIXPublicKey(ca.pksk.Public())
 	if err != nil {
-		t.Fatal(err)
+		t.Fatal("error marshalling public key")
```
whalelines, line 96: Might it be beneficial to include the returned `err` in the fatal log message?

ca/ca_test.go
```diff
@@ -70,19 +78,30 @@ func TestCertCreation(t *testing.T) {
...
+	}
+	testEcPub, err := x509.MarshalPKIXPublicKey(testCa.pksk.Public())
+	if err != nil {
+		t.Fatal("error marshalling test public key")
```
whalelines, line 100: Include `err`?

ca/ca.go
```diff
@@ -160,16 +160,16 @@ func (a *Ca) PktTox509(pktCom []byte, caBytes []byte) ([]byte, error) {
 	}
 
 	var payload struct {
-		Issuer   string `json:"iss"`
-		Audience string `json:"aud"`
-		Email    string `json:"email"`
+		Issuer   string   `json:"iss"`
+		Audience []string `json:"aud"`
```
whalelines, line 164: Why is this changing?

whalelines (commented):
Looks good. Thanks for the tests. A couple questions.

ca/ca.go
mrjoelkamp: everything in `ca.go` is as-is but I had to fix a couple of bugs to make it testable. see https://github.com/openpubkey/openpubkey/blob/dad1e2c7e3838d59e51178f329dce946308ac3e1/client/providers/mocks.go#L17

ca/ca.go
mrjoelkamp: its a bug fix, not sure how it worked before (failing tests) and audience is a list.
[oidc spec](https://openid.net/specs/openid-connect-core-1_0.html):
>aud
REQUIRED. Audience(s) that this ID Token is intended for. It MUST contain the OAuth 2.0 client_id of the Relying Party as an audience value. It MAY also contain identifiers for other audiences. In the general case, the aud value is an array of case-sensitive strings. In the common special case when there is one audience, the aud value MAY be a single case-sensitive string.

ca/ca_test.go
mrjoelkamp: Since this is a test `err` is actually much more useful than the string since the output gives you the line number. Fixed in https://github.com/openpubkey/openpubkey/pull/79/commits/64bfdaf938062a21df12a9b3ac2068d075b7dd6d

ca/ca_test.go
mrjoelkamp: https://github.com/openpubkey/openpubkey/pull/79/commits/64bfdaf938062a21df12a9b3ac2068d075b7dd6d

ca/ca.go
mrjoelkamp: this is better [66550a7](https://github.com/openpubkey/openpubkey/pull/79/commits/66550a72a3300088d08b2c797f91b39e8c4d23cd), trying not to refactor too much 😁

mrjoelkamp: and [e13fc61](https://github.com/openpubkey/openpubkey/pull/79/commits/e13fc617a670233f93c81572d818e3b4d9534150)

ca/ca.go
EthanHeilman: Audience as defined in the oidc spec can be a JSON list, but most OPs I've come across don't represent this as list when it only contains one element. 

Our mock PK Token always treats aud as a JSON list. Our OPs always treat it as a string. If you deserialize it as a list, it works for the unittests but breaks for google. If you treat it as a string, it works for google and breaks for the unittests.

This was how I fixed this issue in PR #68 

```golang
// An audience can be a string or an array of strings.
//
// RFC-7519 JSON Web Token (JWT) says:
// "In the general case, the "aud" value is an array of case-
// sensitive strings, each containing a StringOrURI value.  In the
// special case when the JWT has one audience, the "aud" value MAY be a
// single case-sensitive string containing a StringOrURI value."
// https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.3
var audience string
switch t := claims.Aud.(type) {
case string:
audience = t
case []any:
audList := []string{}
for _, v := range t {
	audList = append(audList, v.(string))
}
audience = strings.Join(audList, ",")
default:
return nil, fmt.Errorf("failed to deserialize aud (audience) claim in ID Token: %d", t)
}
```
https://github.com/openpubkey/openpubkey/pull/68/files#diff-56229f53c5a7b115aa512c3be0077f3e33c4b7b91bc7cc303ecc14f50ddc699bR179-R199

mrjoelkamp: I can add a common `claims` struct type with a custom json unmarshalling method to cover both cases like the code you referenced. Agreed, it is annoying to spec a field that can be two different data types.

mrjoelkamp: what do you think about this? [e7b3e8a](https://github.com/openpubkey/openpubkey/pull/79/commits/e7b3e8a2e7cdf09c7db10a78e68120d0f11359ef)

EthanHeilman: I think a common claims struct as you have done here would be very helpful in a lot of places. Addressing the "Schneiders Audience" issue where you don't know if something is an list of a list until you deserialize it is something I've been planning to fix each time I add unit tests.

This is a really nice way of fixing it and could simplify a lot of code.

mrjoelkamp: Thanks, I had to do something similar recently for unmarshalling the JWKS response while preserving the raw JWK contents. Every OP had a bit different JWK schema, so it was necessary to preserve the original JSON. Especially since some of the JWK claims were not in the standard.

ca/ca.go
```diff
@@ -159,16 +156,12 @@ func (a *Ca) PktTox509(pktCom []byte, caBytes []byte) ([]byte, error) {
...
-		Email    string `json:"email"`
-	}
-	if err := json.Unmarshal(pkt.Payload, &payload); err != nil {
+	payload := new(client.OidcClaims)
+	if err := json.Unmarshal(pkt.Payload, payload); err != nil {
 		return nil, err
 	}
 
-	if payload.Audience != requiredAudience {
+	if payload.Audience != a.reqAud {
```
EthanHeilman, line 164: For github actions ID Tokens, there is no consistent audience to match against since the audience is just CIC commentment.

Instead it might be better to add a Provider in the Ca struct and call

`VerifyPKToken(ctx context.Context, pkt *pktoken.PKToken, provider OpenIdProvider)`

This would get us the same thing and GQ signatures.

EthanHeilman (approved):
I have one comment, but it isn't a blocker as the original CA didn't do any verification of the PK Token. You are welcome to address it or not.

The PR oidcClaim struct this adds will be very nice to have.

ca/ca.go
mrjoelkamp: Agreed, I tried to leave `ca.go` mostly like it was before. I'll update `ca.go` with these changes.

mrjoelkamp: fixed in [82f79c7](https://github.com/openpubkey/openpubkey/pull/79/commits/82f79c7a8322a4b434db2f9ca51004387a5524e7)

EthanHeilman (approved):

EthanHeilman:
Not a code review comment just curious. Is there a name for this pattern you are using here?

```golang
        type Alias OidcClaims
	aux := &struct {
		Audience any `json:"aud"`
		*Alias
	}{
		Alias: (*Alias)(id),
	}

	if err := json.Unmarshal(data, &aux); err != nil {
		return err
	}
```

mrjoelkamp:
> Not a code review comment just curious. Is there a name for this pattern you are using here?

Not sure if there is a name for it but it is necessary to avoid recursive unmarshaling of `OidcClaims` when trying to unmarshal the `aud` field separately in a custom `UnmarshalJSON` function. By casting the receiver (`id *OidcClaims`) to the `Alias` type you can unmarshal the `OidcClaims` struct without calling the custom `UnmarshalJSON` function recursively (since it is now of type `Alias` instead of `OidcClaims`). This embeds the original struct in a new auxiliary struct that you can add additional fields to for separate handling during the unmarshaling process.

PR #86, opened 2024-01-18
Fixes JSON markdown errors in README

EthanHeilman:
The JSON markdown in the README uses some non-JSON formatted strings. Github markdown renders these strings as errors (see picture). This PR removes the JSON tag so that they no longer show up as errors. Also does minor rewording.

Changes this:

image

To this:

image


asamborski (approved):
Thanks for fixing this, Ethan!

PR #88, opened 2024-01-22
Verification Redesign

lgmugnier:
This PR expands the existing `PKTokenVerifier` object into a fully-fledged customizable and parameterized verification functionality. Devs looking for the original object will see this has been removed, however this is unlikely to be felt by users of this library as it was never properly released.

There is a new verifier package that contains this functionality. The `ProviderVerifier` and `CosignerVerifier` allow for custom logic with clearly defined requirements that do not rely on the existence of a `OpenIdProvider`. Because of this logic isolation, the interface has been simplified. It should be noted here that although there was an `Issuer()` function added to the interface, it was unused by the client and easily removed. It was also unreleased.

The `OpenIdProvider` as of the last release was as follows:
```golang
	RequestTokens(ctx context.Context, cicHash string) (*memguard.LockedBuffer, error)
	PublicKey(ctx context.Context, idt []byte) (crypto.PublicKey, error)
	VerifyCICHash(ctx context.Context, idt []byte, expectedCICHash string) error
	VerifyNonGQSig(ctx context.Context, idt []byte, expectedNonce string) error
}
```
and has been changed to
```golang
	RequestTokens(ctx context.Context, cicHash string) (*memguard.LockedBuffer, error)
	PublicKey(ctx context.Context, headers jws.Headers) (crypto.PublicKey, error)
	Verifier() verifier.ProviderVerifier
}
```

This may cause issues for library users that have implemented an object that satisfies the `OpenIdProvider` interface as they will have to create a new function. Although, this is an overall improvement for its simplicity and, in most cases, will reduce the amount of code required to implement a satisfying object. In addition, all verification logic has been consolidated for security and maintainability in the `verifier` package and no longer exists in the PKToken object. Users that have developed custom verifiers relying on these functions, should create an Issue if the new `Verifier` object cannot satisfy the existing purposes.

Two small items have been changed:

1. No impact to library users: the `GQSecurityParameter` has been removed from the `client` package and replaced with a new, helpful GQ SignerVerifier constructor: `New256SignerVerifier`.
2. Impact to library users:`ErrNonGQUnsupported` has been moved from the `client` to the `verifier`

In addition, three functions have been deprecated. Any users of the library may rely on these as they were exported functions in the last release:

1. `VerifyPKToken(ctx context.Context, pkt *pktoken.PKToken, provider OpenIdProvider) error`
has been deprecated in favor of:
```golang
pktVerifier, err := verifier.New(provider.Verifier())
err = pktVerifier.VerifyPKToken(context.Background(), pkt)
```
2. `ExtractClaim(idt []byte, claimName string) (string, error)`
Has been deprecated because it is not used internally.
3. `DiscoverPublicKey(ctx context.Context, headers jws.Headers, issuer string) (crypto.PublicKey, error)`
Has been moved to its own package `discover` to prevent import cycles and its name changed as it no longer has the beneficial context of its location but its accepted and returned args are the same.
```golang
ProviderPublicKey(ctx context.Context, headers jws.Headers, issuer string) (crypto.PublicKey, error)
```


ca/ca.go
```diff
@@ -162,7 +162,7 @@ func (a *Ca) PktTox509(pktCom []byte, caBytes []byte) ([]byte, error) {
 		return nil, err
 	}
 
-	err := client.VerifyPKToken(context.Background(), pkt, a.provider)
+	err := pkt.Verify(context.Background(), a.provider.CommitmentClaim())
```
EthanHeilman, line 165: Why pass the CommitmentClaim()? Could we put the commitment claim in the pheader of the CIC signature?

EthanHeilman: The more I think about your approach here the more I like it. It is nice way to capture the fact that location of the commitment is known by the provider.

EthanHeilman:
@lgmugnier  I'd like to get their PR merged. Moving the verifier to a verifier package would be great. How far away is this from being ready for a code review.

cosigner/cosigner_test.go
```diff
@@ -45,7 +45,7 @@ func TestSimpleCosigner(t *testing.T) {
 	require.NoError(t, err, "failed to generate key pair")
 
 	cosignerClaims := pktoken.CosignerClaims{
-		Iss:         "example.com",
+		Issuer:      "example.com",
```
EthanHeilman, line 48: @lgmugnier Good catch. Mind if I break this out into a separate smaller PR to get it in faster?

examples/ssh/sshcert/sshcert_test.go
```diff
@@ -84,30 +87,38 @@ func TestCASignerCreation(t *testing.T) {
...
-	if err != nil {
-		t.Fatal(err)
+
+	// extra ID token payload claims
+	mockEmail := "arthur.aardvark@example.com"
+	extraClaims := map[string]any{
+		"email": mockEmail,
 	}
 
+	op, err := clientmocks.NewMockOpenIdProvider(t, extraClaims)
```
EthanHeilman, line 102: Nice! I thought about using this pattern when I added email, but decided to do it in a later PR that I never got around to. Glad to see this get added.

You moved this is to clientmocks (client.mocks) to remove some of the cyclical dependency issues with client?

verifier/verifier.go
```diff
@@ -0,0 +1,97 @@
...
+			}
+
+			if claims.Issuer != v.issuer {
+				return fmt.Errorf("expected cosigner: %s, expected %s", claims.Issuer, v.issuer)
+			}
+		}
+	}
+
+	// Enforce all additional, optional checks
+	for _, option := range extraChecks {
```
EthanHeilman, line 89: This is really nice and extendable

verifier/verifier.go
```diff
@@ -0,0 +1,97 @@
...
+
+type verifier struct {
+	issuer           string
+	commitmentClaim  string
+	cosignerVerifier cosignerOpts
+}
+
+func New(issuer, commitmentClaim string) Verifier {
+	return &verifier{
+		issuer:          issuer,
```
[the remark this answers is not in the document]
lgmugnier: > Your motivation here is to not depend on client.provider because the verifier package should be using structs from the client package for reasons of architecture and cyclical dependences?

The Verifier can't use the same interface as the Client because they require different things. Defining an interface for the Verifier would probably look like
```golang
type ProviderVerifier interface {
    Issuer() string
    CommitmentClaim() string
}
```
which is just an unnecessary complex wrapper around two constants.

> New needs more than the issuer to verify the ID Token, we need to know the aud/client-id as well to check it on the ID Token.

> Maybe it makes sense to have some sort of verifier struct/interface that has issuer, audience, and a JWKS resolver function. The JWKS resolver function could then be extended for looking up archival OP pubkeys.

These are good points that I hadn't thought about including, but make sense to have as options.

examples/ssh/sshcert/sshcert_test.go
lgmugnier: I just put it there because I started with the client code. I'm still moving it around to see where it feels best as I modify other tests.

cosigner/cosigner_test.go
lgmugnier: sure, if you'd like. Is it going into a new versioned release? Or do you just want to get it into a different PR?

ca/ca.go
lgmugnier: That was also my thinking but I have been waffling between it being internally provided. I don't like having this extra thing that people need to understand and implement but...

EthanHeilman: > I don't like having this extra thing that people need to understand and implement

I believe you can take this approach and still abstract these details away from most people using OpenPubkey. The verifier can be supplied with OP configuration that includes the commitment claim, then most developers will never have be exposed to this.

```golang
v := verifierpkg.Verifier(verifierpkg.GoogleOP)

err:= v.VerifyPKToken(pkt)
```

Under the hood this will be calling `pkt.Verify(context.Background(), a.provider.CommitmentClaim())` but the developer doesn't need to see those details unless they are really mucking about and writing their own OP configs.

verifier/verifier.go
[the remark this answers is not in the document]
EthanHeilman: > The Verifier can't use the same interface as the Client because they require different things

What about having a common OP struct that gets different parts used by two different interfaces: a client interface and a verifier interface. The advantage of that approach is someone adding a new provider only has to create one file and all the information about that provider is stored together.

> which is just an unnecessary complex wrapper around two constants and a bit awkward to make configurable.

With Google you need at least three constants: the issuer, the audience and the commitment claim. If we added gitlab we would need the ability to tell it to not have a commitment claim and add the CIC via the GQ signature. So I guess what I'm saying here is that it would be nice to have a per OP function which is called that makes this decisions in a flexible and programmable way.

> These are good points that I hadn't thought about including, but make sense to have as options.

Agreed and I believe that was the motivation behind having the PublicKey() method for OpenID Providers. You can use a decorator or just implement a new provider on the interface to point it  to an archival JWKS.

cosigner/cosigner_test.go
EthanHeilman: I just broke it out into it's own, very small PR.

Normally I was say this should get a versioned release because it breaks downstream, but we are lucky in that no one depends on the Cosigner code yet.
https://github.com/openpubkey/openpubkey/pull/99

client/opkclient_test.go
```diff
@@ -38,36 +37,38 @@ func TestClient(t *testing.T) {
...
 				require.Equal(t, signer, c.GetSigner(), tc.name)
-				require.Equal(t, tc.alg, c.GetAlg(), tc.name)
+				require.Equal(t, tc.signerAlg, c.GetAlg(), tc.name)
 			} else if tc.gq {
-				c, err = client.New(op, client.WithSignGQ(tc.gq))
+				c, err = client.New(provider, client.WithSignGQ(tc.gq))
 				require.NoError(t, err, tc.name)
 			} else {
-				c, err = client.New(op)
+				c, err = client.New(provider)
```
EthanHeilman, line 71: I like your rename of signerAlg here as it increases clarity. My original name was a little confusing, is this the alg for the OP or the alg for the signer.

For the same reasons I like you renaming of signerAlg, I find the renaming of op to provider to be confusing since we have OpenIDProviders and CosignerProviders and the client can take both of them.

You know me, I love short names so I prefer op (OpenIDProvider) and cosP (CosignerProvider), but you really must to use a longer name here, can you use a name that makes it clear this is an OpenIDProvider such as `oidcProvider`, `oProvider`. Just something that allows a develop at a glance to know which provider you are referencing.

client/opkclient_test.go
```diff
@@ -104,43 +105,30 @@ func TestClient(t *testing.T) {
...
-				require.NoError(t, err, tc.name)
-
-				sv, err := gq.NewSignerVerifier(opPubKey.(*rsa.PublicKey), client.GQSecurityParameter)
-				require.NoError(t, err, tc.name)
-
-				ok := sv.VerifyJWT(idt)
-				if !ok {
-					t.Fatal(fmt.Errorf("error verifying OP GQ signature on PK Token (ID Token invalid)"))
-				}
+				require.Equal(t, gq.GQ256, providerKeyAlgorithm, tc.name)
```
EthanHeilman, line 128: What is your motivation behind removing the test that client produced a valid GQ signature?

EthanHeilman:
Very excited to see this is ready for review! I'm making some minor comments right now, let me do a deeper dive tomorrow.

client/opkclient_test.go
EthanHeilman: Reading more of the code and I think I see what you are doing here. You have renamed the CosignerProvider to Cosigner so that provider always implies OP. Let me get that some thought. I think you can ignore my above comment for now, I wrote it before I got to code where you make that change.

client/opkclient_test.go
lgmugnier: I think it was to reduce test redundancy particularly when we don't do the same for other algorithms. This should be tested (which it is) in the verifier package. The client tester should focus on testing the inputs and outputs of the client functions which is why it makes sense to check if the pk token has a certain algorithm. 

I also think we've encountered an issue in previous tests where the code used to test wasn't accurate or was assuming too much and this helps to keep everything tight and reliable.

client/opkclient_test.go
lgmugnier: Go really wants you to use single, whole words, a point you've brought up in the past, because it's a lot more readable and explanatory. I will say that I find it confusing to have an oidc provider and a cosigner provider so assigning each a distinct yet descriptive word I think helps with the overall clarity. But I'll wait for your further comments.

verifier/cosignerverifier.go
```diff
@@ -0,0 +1,94 @@
...
+	issuer  string
+	options CosignerVerifierOpts
+}
+
+type CosignerVerifierOpts struct {
+	// Strict specifies whether or not a pk token MUST contain a signature by this cosigner.
+	// Defaults to false.
+	Strict bool
+	// Allows users to set custom function for discovering public key of Cosigner
+	DiscoverPublicKey func(ctx context.Context, kid string, issuer string) (jwk.Key, error)
```
EthanHeilman, line 26: This pattern here is really nice. I can't wait to use it to plug my historical public key package into cosigner.

verifier/verifier.go
```diff
@@ -0,0 +1,114 @@
...
+			}
+
+			// Verify cosigner signature
+			if err := cosignerVerifier.VerifyCosigner(ctx, pkt); err != nil {
+				return err
+			}
+
+			// If any other cosigner verifiers are set to strict but aren't present, then return error
+			for _, cosignerVerifier := range v.cosigners {
+				if cosignerVerifier.options.Strict && cosignerVerifier.issuer != cosignerClaims.Issuer {
```
EthanHeilman, line 80: If we have two or more cosignerVerifier with `options.Strict=true` won't this always fail?

If we view the cosignerVerifier list as a filter, it needs to match on a cosignerVerifier. Do we need `options.Strict` at all?

verifier/verifier.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/lestrrat-go/jwx/v2/jws"
+	"github.com/openpubkey/openpubkey/pktoken"
+)
+
+type Verifier struct {
+	providers map[string]*ProviderVerifier
+	cosigners map[string]*CosignerVerifier
+}
+
+func New(verifier *ProviderVerifier, options ...VerifierOpts) *Verifier {
```
EthanHeilman, line 16: If only one verifier can be specified why is providers a list? How do you specify more than one provider?

verifier/providerverifier.go
```diff
@@ -0,0 +1,280 @@
...
+type ProviderVerifier struct {
+	issuer          string
+	commitmentClaim string
+	options         ProviderVerifierOpts
+}
+
+type ProviderVerifierOpts struct {
+	// If ClientID is specified, then verification will require that the ClientID
+	// be present in the audience ("aud") claim of the PK token payload
+	ClientID string
```
EthanHeilman, line 28: Given the how common and how damaging failure to check `aud=client-id` has been in the OIDC ecosystem, we should not default to not checking the `aud=client-id`. For particular OPs like Google the audience should be required and we should not allow a developer to turn it off without writing a custom OP.

verifier/providerverifier.go
```diff
@@ -0,0 +1,280 @@
...
+	DiscoverPublicKey func(ctx context.Context, kid string, issuer string) (jwk.Key, error)
+	// Allows for successful verification of expired tokens
+	SkipExpirationCheck bool
+}
+
+// Creates a new ProviderVerifier with required fields
+//
+// issuer: Is the OpenID provider issuer as seen in ID token e.g. "https://accounts.google.com"
+// commitmentClaim: the ID token payload claim name where the cicHash was stored during issuance
+func NewProviderVerifier(issuer, commitmentClaim string, options ProviderVerifierOpts) *ProviderVerifier {
```
EthanHeilman, line 39: We should provide pre-configured ProviderVerifier for at least Google and Github, similar to how we provide client.providers.Google. This would let us hide the commitmentClaim.

examples/mfa/example.go
```diff
@@ -77,13 +78,9 @@ func main() {
...
 
 		// Verify our pktoken including the cosigner signature
-		verifier := client.PKTokenVerifier{
-			AllowedProviders: []client.OpenIdProvider{provider},
-			AllowedCosigners: []client.CosignerProvider{cosignerProvider},
-		}
-
-		// Verify our pktoken including the cosigner signature
-		if err := verifier.Verify(context.TODO(), pkt); err != nil {
+		providerVerifier := verifier.NewProviderVerifier(provider.Issuer(), provider.CommitmentClaim(), verifier.ProviderVerifierOpts{})
```
EthanHeilman, line 81: If I'm reading this correctly, this introduces a vulnerability in that it removes the  `aud=client-id` check.

EthanHeilman: This supports the idea that we should have a GoogleVerifier that requires the `aud=client-id` check.

verifier/verifier.go
lgmugnier: I was hoping to make this forward thinking. We've always said we wanted multiple cosigners, this is something we can accommodate at this stage

verifier/verifier.go
lgmugnier: Using the Verifier options accepted in the "New" function

verifier/providerverifier.go
lgmugnier: [0d9c454](https://github.com/openpubkey/openpubkey/pull/88/commits/0d9c454585fe3122a692e2ead0e6031f100b8bb0)

client/opkclient_test.go
EthanHeilman: If we don't do the same for the other algorithm then we should add a test for the other algorithms rather than remove the test.

pktoken/pktoken.go
```diff
@@ -53,12 +54,8 @@ type PKToken struct {
 
 // kid isn't always present, and is only guaranteed to be unique within a given key set,
 // so we can use the thumbprint of the key instead to identify it at verification time
-func (p *PKToken) AddJKTHeader(opKey crypto.PublicKey) error {
-	public, err := jwk.FromRaw(opKey)
-	if err != nil {
-		return fmt.Errorf("failed to create JWK from public key: %w", err)
-	}
-	thumbprint, err := public.Thumbprint(crypto.SHA256)
+func (p *PKToken) AddJKTHeader(opKey jwk.Key) error {
```
EthanHeilman, line 57: Does this change impact downstream projects?

verifier/cosignerverifier.go
```diff
@@ -0,0 +1,94 @@
...
+
+type CosignerVerifier struct {
+	issuer  string
+	options CosignerVerifierOpts
+}
+
+type CosignerVerifierOpts struct {
+	// Strict specifies whether or not a pk token MUST contain a signature by this cosigner.
+	// Defaults to false.
+	Strict bool
```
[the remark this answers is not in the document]
lgmugnier: 76fe1d9

pktoken/pktoken.go
EthanHeilman: The typical set of downstream projects I check are:
* All the projects in OpenPubkey org
* then the sshpiper code

In this case Docker added this specifically for something they were building, but I don't see where they use it so the code this depends on probably isn't public.

This change is seems out of scope for this PR. If there some compelling reason to make this change (Is there a bug I'm missing?), open a seperate PR and make the case in that PR so we can manage the downstream dependencies independently.

lgmugnier: I have changed it back [bd5d2f4](https://github.com/openpubkey/openpubkey/pull/88/commits/bd5d2f4aae0929b6db358df793330ae4982dd9dd)

However, if this is a requirement of PRs going forward, then you should add it to the agenda of either the community or TSC meeting.

verifier/verifier_test.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+	pkt, err = opkClient.Auth(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	providerVerifier = verifier.NewProviderVerifier(provider.Verifier().Issuer(), commitmentClaim, verifier.ProviderVerifierOpts{ClientID: clientID})
+	pktVerifier = verifier.New(providerVerifier)
+	err = pktVerifier.VerifyPKToken(context.Background(), pkt, verifier.GQOnly())
+	if err != nil {
```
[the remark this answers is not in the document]
lgmugnier: ecb6915

client/client.go
```diff
@@ -26,21 +26,36 @@ import (
...
 	"github.com/openpubkey/openpubkey/pktoken/clientinstance"
 	"github.com/openpubkey/openpubkey/util"
+	"github.com/openpubkey/openpubkey/verifier"
 )
 
-// OpkClient is the OpenPubkey client
+// Interface for interacting with the OP (OpenID Provider)
+type OpenIdProvider interface {
+	RequestTokens(ctx context.Context, cicHash string) (*memguard.LockedBuffer, error)
+	Verifier() verifier.ProviderVerifier
```
EthanHeilman, line 40: Is there a reason not to use composition for the Verifier here? That way rather than `op.Verifier().Issuer()` you could just call `op.Issuer()`?

verifier/providerverifier.go
```diff
@@ -0,0 +1,305 @@
...
+	if err := v.verifyCommitment(pkt); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// This function takes in an OIDC Provider created ID token or GQ-signed modification of one and returns
+// the associated public key
+func (v *DefaultProviderVerifier) ProviderPublicKey(ctx context.Context, token []byte) (jwk.Key, error) {
```
EthanHeilman, line 107: Given the name for this function was `PublicKey` and the fact that this is a Provider is implied by the fact that this is a `ProviderVerifier`, I don't see the benefit of renaming this.

Why change the returned type from `crypto.PublicKey` to `jwk.Key`. crypto.PublicKey is a standard part of the golang crypto library whereas changing this to `jwk.Key` makes our interface dependent on the jwx library. A dependency on the `jwk.Key` in the function signature means switching libraries under the hood is more diffcult.

A while ago you pushed for a change to use `crypto.signer` in all function signatures and that made our function signatures much nicer for the same reason.

I'm not advocating for changing functions that have jwx object types in the function arguments but we shouldn't make the problem worse.

client/opkclient_test.go
EthanHeilman: I think I misunderstood your intention here. I thought you were renaming op to provider everywhere because you were no longer referring to the cosigner as a provider. However it appears the cosigner is still named a cosigner provider.  Given that we use the Op in the client struct and in the constructor to the client `func New(op OpenIdProvider, opts ...ClientOpts) (*OpkClient, error) {` I don't see a justification to renaming and touching all these lines of code in the tests.

This is existing code and changing it is both out of scope for this PR and reduces the consistency of variable names in this code base.

verifier/verifier_test.go
```diff
@@ -0,0 +1,126 @@
...
+	require.NoError(t, err)
+
+	// Check if verification fails with incorrect issuer
+	wrongIssuer := "https://evil.com/"
+	providerVerifier := verifier.NewProviderVerifier(wrongIssuer, commitmentClaim, verifier.ProviderVerifierOpts{SkipClientIDCheck: true})
+	pktVerifier = verifier.New(providerVerifier)
+	err = pktVerifier.VerifyPKToken(context.Background(), pkt)
+	require.Error(t, err)
+
+	// Check if verification failes with incorrect commitment claim
```
EthanHeilman, line 42:
```suggestion
	// Check if verification fails with incorrect commitment claim
```

EthanHeilman (changes requested):

client/client.go
```diff
@@ -26,21 +26,36 @@ import (
...
-	cosP   *CosignerProvider
-	signer crypto.Signer
-	alg    jwa.KeyAlgorithm
-	signGQ bool // Default is false
+	Op       OpenIdProvider
+	cosP     *CosignerProvider
+	signer   crypto.Signer
+	alg      jwa.KeyAlgorithm
+	signGQ   bool // Default is false
+	verifier PKTokenVerifier
```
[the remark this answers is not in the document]
lgmugnier: The OP has all of the knowledge required to verify the OP, that's why the OP's verifier returns a `ProviderVerifier`, it does not have knowledge about cosigner verification and should not. 

The PKTokenVerifier is there to represent an entire verifier, this allows users who may have general verifiers to use that instead of something more specific to this single PK token. This is entirely for flexibility and I do not see it as something that is easily misconfigured unless someone is explicitly looking to use it.

client/client.go
lgmugnier: `Issuer()` is required by the verifier not the client. That's why `op.Verifier().Issuer()` is not used in the code base, just tests for test reasons and not for real-world reasons.

verifier/verifier.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+	}
+}
+
+type Verifier struct {
+	providers map[string]ProviderVerifier
+	cosigners map[string]CosignerVerifier
+}
+
+func New(verifier ProviderVerifier, options ...VerifierOpts) *Verifier {
```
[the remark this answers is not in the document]
lgmugnier: My intention with the constructor in this case was to clearly delineate between what was required and what is optional. At a minimum (as reflected in the constructor) we require a single `ProviderVerifier`. The `VerifierOpts` are exactly that: options. If you would like, you can add more providers and specify cosigners as you've shown.

client/opkclient_test.go
```diff
@@ -103,43 +103,31 @@ func TestClient(t *testing.T) {
...
 
-			thumbprint, err := pub.Thumbprint(crypto.SHA256)
+			thumbprint, err := pubkey.Thumbprint(crypto.SHA256)
 			require.NoError(t, err, tc.name)
 
 			thumbprintStr := string(util.Base64EncodeForJWT(thumbprint))
 			require.Equal(t, jktstr, thumbprintStr, "jkt header does not match op thumbprint in "+tc.name)
 
-			alg, ok := pkt.ProviderAlgorithm()
+			providerKeyAlgorithm, ok := pkt.ProviderAlgorithm()
```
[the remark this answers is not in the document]
lgmugnier: There are many instances in test files where we have to use different names besides `alg` because there are multiple variables to which that name would apply. That is also the case here. I was trying to complement your naming choice in `signerAlg`. I can shorten it to `providerAlg` which I think would make sense, but since you don't suggest that, what do you suggest I do?

client/claims.go
```diff
@@ -0,0 +1,66 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package client
```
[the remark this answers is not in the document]
lgmugnier: I often struggle to remember that the interfaces are in this file with a bunch of other seemingly random things. I moved the relevant interfaces to the client file because they were significantly changed and it was a good opportunity to put them in the correct place (according to Go). I would love to delete the file entirely but I kept it around for legacy purposes as it doesn't contain any code that we use in our library.

I think it is important to also prioritize new developers coming in and where they might look for something like this. Someone who is not constantly touching the code may likely go to default expected locations (like me) as opposed to where they "remember" it being.

verifier/providerverifier.go
lgmugnier: So, I found this to be a significantly important change. I do agree that relying on the jwx libraries in functional signatures is not ideal, and I'm happy to create our own object that alias's it. However, we should be capturing the entire key including any options it may contain and not just the public key. I know that we're not really doing anything with it now but it seems unfortunate short-term thinking to be creating a pattern that ignores part of the jwks.

verifier/verifier.go
```diff
@@ -0,0 +1,147 @@
...
+			fmt.Println(verifier.issuer)
+			v.cosigners[verifier.issuer] = verifier
+		}
+	}
+}
+
+func AddProviderVerifiers(verifiers ...ProviderVerifier) VerifierOpts {
+	return func(v *Verifier) {
+		for _, verifier := range verifiers {
+			v.providers[verifier.Issuer()] = verifier
```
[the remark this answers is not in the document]
lgmugnier: [53cab2e](https://github.com/openpubkey/openpubkey/pull/88/commits/53cab2ebe1938b6542aee76798fe824cd2e3f9f5)

client/client.go
EthanHeilman: The issuer value is required by the client. It needs to the issuer value to look up 


```golang
var (
	key            = []byte("NotASecureKey123")
	issuedAtOffset = 1 * time.Minute
)
const googleIssuer = "https://accounts.google.com"
type GoogleOp struct {
	ClientID        string
```
https://github.com/openpubkey/openpubkey/pull/88/files#diff-d18f44e884f4688a7c606a2716462f72a73686038059b03101a7e72cbcb04307L44

The issuance code in the Op needs to the issuer URI to do the well-known look up and get the authorization_endpoint to initiate authorization.

client/client.go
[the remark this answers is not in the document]
EthanHeilman: I agree the OP should and does not have the information to verify the cosigner. We are discussing the client here which if it is handling also getting the PK Token cosigned, should have a cosigner object which can verify the cosigned PK Token.

> This is entirely for flexibility and I do not see it as something that is easily misconfigured unless someone is explicitly looking to use it.

The only use for this seems to be to avoid changing the verification logic in the VerifierProvider in the OP? But shouldn't we be encouraging developers to add the customer verification logic in the VerifierProvider not in multi OP Verifier?

What is the usecase where it would be better for a user to put custom verification logic in the multi OP Verifier than in the VerifierProvider in the OP?

client/opkclient_test.go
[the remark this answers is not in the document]
EthanHeilman: Either `providerAlg` or `opAlg` works for me. it is improvement in readability.

The function `ProviderAlgorithm` on the PK Token confusing because we have OpenIDProviders and CosignerProviders, but given that that the name `ProviderAlgorithm` is not a change in this PR, I would not advocate for a rename. We might just want to move away from calling the CosignerProvider a provider.

verifier/providerverifier.go
EthanHeilman: > However, we should be capturing the entire key including any options it may contain and not just the public key.

Is the main option you are concerned here alg?

verifier/providerverifier.go
```diff
@@ -0,0 +1,305 @@
...
+	if err := v.verifyCommitment(pkt); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// This function takes in an OIDC Provider created ID token or GQ-signed modification of one and returns
+// the associated public key
+func (v *DefaultProviderVerifier) ProviderPublicKey(ctx context.Context, token []byte) (JSONWebKey, error) {
```
EthanHeilman, line 107: This PR changes `func DiscoverPublicKey(ctx context.Context, headers jws.Headers, issuer string) (crypto.PublicKey, error)`  to  `func (v *DefaultProviderVerifier) ProviderPublicKey(ctx context.Context, token []byte) (JSONWebKey, error)`

This change breaks the future ability to look up a PublicKey based on a `jtk` public header claim as the new ProviderPublicKey function takes a compact Token. Compact Tokens exclude public header claims. The `jtk` public header functionality was added specifically to enable the ability to look up an OP Public Key when the OP JWKS has no kid (see https://github.com/openpubkey/openpubkey/pull/73 )

This is also used by a our GQ Signature header format:
https://github.com/openpubkey/openpubkey/pull/66/files

In the current codebase, we throw an error if the kid does not exist rather than falling back to the jtk claim in the headers, but when we add a provider that doesn't have a `kid` we can add support for this. 
```
func DiscoverPublicKey(ctx context.Context, headers jws.Headers, issuer string) (crypto.PublicKey, error) {
    discConf, err := oidcclient.Discover(issuer, http.DefaultClient)
    if err != nil {
        return nil, fmt.Errorf("failed to call OIDC discovery endpoint: %w", err)
    }


    jwks, err := jwk.Fetch(ctx, discConf.JwksURI)
    if err != nil {
        return nil, fmt.Errorf("failed to fetch to JWKS: %w", err)
    }


    kid := headers.KeyID()
    key, ok := jwks.LookupKeyID(kid)
    if !ok {
        return nil, fmt.Errorf("key %q isn't in JWKS", kid)
    }
```
By changing the PublicKey on an OP and Verifier to no longer accept headers we lose the flexibility to add feature without changing the OpenIdProvider and VerifierProvider interface.

Suggested Fix:  We've already talking about reverting the changes to the OpenIdProvider  interface so that should address this as well. I just want to leave a note here to remember to check that we didn't break the path to look up PublicKeys by jtk claims.

client/providers/github_actions.go
```diff
@@ -68,25 +67,8 @@ func buildTokenURL(rawTokenURL, audience string) (string, error) {
...
-}
-
-func (g *GithubOp) Issuer() string {
-	return githubIssuer
-}
-
-func (g *GithubOp) PublicKey(ctx context.Context, headers jws.Headers) (crypto.PublicKey, error) {
-	return client.DiscoverPublicKey(ctx, headers, githubIssuer)
+func (g *GithubOp) Verifier() verifier.ProviderVerifier {
+	return verifier.NewProviderVerifier(githubIssuer, "aud", verifier.ProviderVerifierOpts{})
```
EthanHeilman, line 71: Github fail to verify a non-GQ PK Token as is the current behavior. Does this change that behavior?

```golang
func (*GithubOp) VerifyNonGQSig(context.Context, []byte, string) error {
	return client.ErrNonGQUnsupported
}
```
https://github.com/openpubkey/openpubkey/blob/main/client/providers/github_actions.go#L130C1-L132C2

client/opkclient_test.go
```diff
@@ -38,31 +37,33 @@ func TestClient(t *testing.T) {
...
+		{name: "with GQ, with signer, with extraClaims", gq: true, signer: true, signerAlg: jwa.ES256, extraClaims: map[string]string{"extra": "yes"}},
 		{name: "with GQ, with extraClaims", gq: true, signer: false, extraClaims: map[string]string{"extra": "yes", "aaa": "bbb"}},
 	}
 
-	op, err := providers.NewMockOpenIdProvider()
-	require.NoError(t, err, "failed to create mock OpenIdProvider")
+	op, err := mocks.NewMockOpenIdProvider(t, map[string]any{})
+	if err != nil {
+		t.Fatal(err)
+	}
```
EthanHeilman, line 54:
```suggestion
require.NoError(t, err, "failed to create mock OpenIdProvider")
```

client/providers/mocks/oidp_server.go
```diff
@@ -0,0 +1,123 @@
...
+
+type OIDPServer struct {
+	listener  net.Listener
+	Server    *http.ServeMux
+	uri       string
+	kid       string
+	jwksBytes []byte
+}
+
+// A very simple JWKS server for our MFA Cosigner example code.
```
EthanHeilman, line 45:
```suggestion
// A very simple JWKS server for setting up a test OpenID Provider
```

client/providers/mocks/provider.go
```diff
@@ -0,0 +1,146 @@
...
+	return r0
+}
+
+// RequestTokens provides a mock function with given fields: ctx, cicHash
+func (_m *OpenIdProvider) RequestTokens(ctx context.Context, cicHash string) (*memguard.LockedBuffer, error) {
+	ret := _m.Called(ctx, cicHash)
+
+	var r0 *memguard.LockedBuffer
+	var r1 error
+	if rf, ok := ret.Get(0).(func(context.Context, string) (*memguard.LockedBuffer, error)); ok {
```
EthanHeilman, line 46: I am a little confused by these variable names. Why is r0 a LockedBuffer, but r1 is an error, what is rf? What does the r stand for?

client/providers/mocks/oidp_server.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/lestrrat-go/jwx/v2/jwk"
+	"github.com/zitadel/oidc/v2/pkg/oidc"
+)
+
+const (
+	jwksEndpoint            = "/.well-known/jwks.json"
+	wellKnownConfigEndpoint = "/.well-known/openid-configuration"
+)
+
+type OIDPServer struct {
```
EthanHeilman, line 37: This looks like the JWKS Server from my MFA example code.  Name seems a little confusing because it just serves well-known configuration and JWKS isn't an OP. Maybe call it `JwksServer` or `JWKSServer` unless the plan is to add more OP functionality to it.

Not a blocker

client/providers/mocks/provider.go
EthanHeilman: Opps, I miss the autogenerated mock type comment at the type. Going to resolve this.

client/providers/mocks/provider.go
```diff
@@ -0,0 +1,146 @@
...
+	"time"
+
+	"github.com/awnumar/memguard"
+	"github.com/lestrrat-go/jwx/v2/jwa"
+	"github.com/lestrrat-go/jwx/v2/jws"
+	"github.com/openpubkey/openpubkey/verifier"
+	"github.com/stretchr/testify/mock"
+)
+
+// OpenIdProvider is an autogenerated mock type for the OpenIdProvider type
```
EthanHeilman, line 19: What tool/command did you use to generate this? A comment describing how to generate this would be useful if anyone needs to regenerate this code

examples/ssh/opkssh/cli_test.go
```diff
@@ -18,24 +21,30 @@ func AllowAllPolicyEnforcer(userDesired string, pkt *pktoken.PKToken) error {
...
+	// extra ID token payload claims
+	mockEmail := "arthur.aardvark@example.com"
+	extraClaims := map[string]any{
+		"email": mockEmail,
+	}
+
+	op, err := providermocks.NewMockOpenIdProvider(t, extraClaims)
 	if err != nil {
 		t.Fatal(err)
 	}
```
EthanHeilman, line 39: Not a issue you introduced, but since you are editing the code on either side, it is worth fixing

```suggestion
require.NoError(t, err)
```

client/providers/github_actions.go
EthanHeilman: I see you have the GQOnly() check. I think you can just add that to the github ProviderVerifier

verifier/verifier.go
```diff
@@ -0,0 +1,153 @@
...
+	Strict() bool // Whether or not a given cosigner MUST be present for successful verification
+	VerifyCosigner(ctx context.Context, pkt *pktoken.PKToken) error
+}
+
+type VerifierOpts func(*Verifier) error
+
+func WithCosignerVerifiers(verifiers ...*DefaultCosignerVerifier) VerifierOpts {
+	return func(v *Verifier) error {
+		for _, verifier := range verifiers {
+			fmt.Println(verifier.issuer)
```
EthanHeilman, line 29: If we are checking for existing issuers in `AddProviderVerifiers` we should add that check here as well

verifier/verifier.go
```diff
@@ -0,0 +1,153 @@
...
+	ctx context.Context,
+	pkt *pktoken.PKToken,
+	extraChecks ...Check,
+) error {
+	issuer, err := pkt.Issuer()
+	if err != nil {
+		return err
+	}
+
+	providerVerfier, ok := v.providers[issuer]
```
EthanHeilman, line 83: Can you add a comment similar to my comment here which specifies the current limitation of providers with different audiences and the same issuer. I know your check above would cause an error but we do eventually want to support the same issuer with different audiences

https://github.com/openpubkey/openpubkey/blob/main/client/oidc.go#L125C1-L133C1

client/client.go
lgmugnier: I'm not sure what you're pointing to here, but there is no location in the client that requires this information. I think you might be mistakenly referring to the Google Provider which calls Issuer() unnecessarily since issuer is essentially a class variable and there is no value in calling the function instead.

README.md
```diff
@@ -52,7 +52,8 @@ signedMsg, err := pkt.NewSignedMessage(msg, opkClient.GetSigner())
 To verify a signed message, we first verify that the PK Token `pkt` is issued by the OP (Google). Then we use the PK Token to verify the signed message.
 
 ```golang
-err = client.VerifyPKToken(context.Background(), pkt, op)
+pktVerifier, err := verifier.New(provider.Verifier())
+err = pktVerifier.VerifyPKToken(context.Background(), pkt)
 msg, err := pkt.VerifySignedMessage(osm)
```
EthanHeilman, line 57: This part of the readme has a minor mistake in it. Didn't notice it until now. Not caused by your PR but worth fixing since you are changing the lines above it.

```suggestion
msg, err := pkt.VerifySignedMessage(signedMsg)
```

verifier/providerverifier.go
```diff
@@ -0,0 +1,277 @@
...
+	}
+	expectedCommitment, err := cic.Hash()
+	if err != nil {
+		return err
+	}
+	commitment, ok := claims[v.commitmentClaim]
+	if !ok {
+		return fmt.Errorf("missing commitment claim %s", v.commitmentClaim)
+	}
+
```
EthanHeilman, line 171: Added a clarifying comment here.

```suggestion

// The OIDC Spec allows the aud (audience) claim to be list or a string. 
// In the case of an aud commitment claim, we assume it is a string and 
// we do not handle the case where it is a list.
```

We might want to add a check if we get a list for the aud commitment and output an error message explaining we encountered an aud commitment claim that is a list. Not a blocker, can be done in a future PR.

verifier/providerverifier.go
```diff
@@ -0,0 +1,277 @@
...
+	if err := v.verifyCommitment(pkt); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// This function takes in an OIDC Provider created ID token or GQ-signed modification of one and returns
+// the associated public key
+func (v *DefaultProviderVerifier) ProviderPublicKey(ctx context.Context, token []byte) (crypto.PublicKey, error) {
```
EthanHeilman, line 112: This would prevent someone from doing an OP Public Key look up based on jkt

EthanHeilman (approved):
LTGM

PR #93, opened 2024-01-31
Creates OpkClient constructor and moves singer/alg/gqflag into the OpkClient struct

EthanHeilman:
The main change is to take three of the five parameters to `client.Auth(...)` and move them into the client struct.

This the current definition of Auth:
``` golang
func (o *OpkClient) Auth(ctx context.Context, signer crypto.Signer, alg jwa.KeyAlgorithm, extraClaims map[string]any, signGQ bool) (*pktoken.PKToken, error) 
```

This is the definition of Auth after this PR:
``` golang
func (o *OpkClient) Auth(ctx context.Context, opts ...AuthOpts) (*pktoken.PKToken, error)
```

### OpkClient constructor

We then create a OpkClient  constructor that supports optional parameters via a variadic function. This lets us initialize these values with defaults but provides flexibility if developers want to override these defaults.

Creating an opkClient, all default values
```golang
opkClient, err := client.New(op)
```

Creating an opkClient overriding all default values
```golan
opkClient, err := client.New(
        op,
	client.WithCosignerProvider(&cosignerProvider),
	client.WithSigner(signer, alg),
	client.WithSignGQ(true))
```

## TODOs

- [X] Transform signer/alg/signGQ parameters in the `client.Auth` parameters to struct variables.
- [X] Moves the client unittest to use `client.Auth` rather than `client.OidcAuth`
- [X] Updates examples to use `client.Auth`
- [X] ExtraClaims as options list in `client.Auth`
- [X] OpkClient.New constructor with getters and setters
- [X] Tests for optional parameter passing to the constructor and auth
- [X] Package docs updates

### Signer/Alg

The signer will be used by for the Cosigner Refresh and POP Auth client functions so it makes sense to bundle them as instance variables of the client. Additionally allows passing of the client into functions without also having to pass the signer and alg variables separately. For example look at how putting the signer and the alg into the struct cleans up the SSH example:

Before
```golang

client := &client.OpkClient{
     Op: &op,
}
gqFalse := false
certBytes, seckeySshPem, err := createSSHCert(context.Background(), client, signer, alg, gqFalse, principals)
...
func createSSHCert(cxt context.Context, client *client.OpkClient, signer crypto.Signer, alg jwa.KeyAlgorithm, gqFlag bool, principals []string) ([]byte, []byte, error) {
	pkt, err := client.Auth(cxt, signer, alg, map[string]any{}, gqFlag)
```

After
```golang
opkClient, err := client.New(
	&op,
	client.WithSigner(signer, alg),
	client.WithSignGQ(false),
)
certBytes, seckeySshPem, err := createSSHCert(context.Background(), opkClient, principals)
...
func createSSHCert(cxt context.Context, client *client.OpkClient, principals []string) ([]byte, []byte, error) {
	pkt, err := client.Auth(cxt)
```
This reduces the number of moving pieces and makes code written using the client simpler, more concise and removes the possibility bugs where another signer is substituted as the client signer.

This pattern allows users to bring their own signer using `WithSigner()`, but if they don't specify a signer we create a signer for them. This means that someone writing a simple app using OpenPubkey has to write much less boilerplate.

Before
```golang
opk := &client.OpkClient{
	Op:   &op
}
clientKey, err := util.GenKeyPair(jwa.ES256)
pkt, err := opk.Auth(context.TODO(), clientKey, jwa.ES256, map[string]any{}, false)
```

After
```
opkClient, err := client.New(&op)
pkt, err := opk.Auth(context.TODO())
```

This also enables login/logout functionality when a client can be torn down and the signer deleted.

### SignGQ

This follows the same justification as Signer/Alg. A client shouldn't want to mix GQ and non-GQ tokens. Configuring this up front when creating the client provides a single place to check how the client is configured. Someone wanted to audit that `SignGQ=true` only needs to check client creation, not each code path where Auth is called.

 ## Backwards compatibility?

No one is using the `client.Auth` method currently as it was just introduced in the last PR a few days ago. We can make this change with or without `client.New` without breaking anything downstream.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -92,7 +92,7 @@ func New(op OpenIdProvider, opts ...ClientOpts) (*OpkClient, error) {
 	}
 
 	if client.alg == nil && client.signer != nil {
-		return nil, fmt.Errorf("signer specified but alg is nil, must specify alg  of signer")
+		return nil, fmt.Errorf("signer specified but alg is nil, must specify alg of signer")
 	}
 
 	if client.signer == nil {
--- a/client/client.go
+++ b/client/client.go
@@ -35,12 +35,100 @@ import (
 )
 
 type OpkClient struct {
-	Op          OpenIdProvider
-	CosP        *CosignerProvider
-	Signer      crypto.Signer
-	Alg         jwa.KeyAlgorithm
-	SignGQ      bool // Default is false
-	ExtraClaims map[string]any
+	Op     OpenIdProvider
+	CosP   *CosignerProvider
+	signer crypto.Signer
+	alg    jwa.KeyAlgorithm
+	signGQ bool // Default is false
+}
+
+// ClientOpts contains options for constructing an OpkClient
+type ClientOpts func(o *OpkClient)
+
+// WithSigner allows the caller to inject their own signer and algorithm.
+// Use this option if to generate to bring your own user key pair. If this
+// option is not set the OpkClient constructor will automatically generate
+// a signer, i.e., key pair.
+// Example use:
+//
+//	signer, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+//	WithSigner(signer, jwa.ES256)
+func WithSigner(signer crypto.Signer, alg jwa.KeyAlgorithm) ClientOpts {
+	return func(o *OpkClient) {
+		o.signer = signer
+		o.alg = alg
+	}
+}
+
+// WithSignGQ specifies if the client should replace the ID Token's signature
+// with a GQ Signature.
+func WithSignGQ(signGQ bool) ClientOpts {
+	return func(o *OpkClient) {
+		o.signGQ = signGQ
+	}
+}
+
+// WithCosignerProvider specifies what cosigner provider should be used to
+// cosign the PK Token. If this is not specified then the cosigning setup
+// is skipped.
+func WithCosignerProvider(cosP *CosignerProvider) ClientOpts {
+	return func(o *OpkClient) {
+		o.CosP = cosP
+	}
+}
+
+func New(op OpenIdProvider, opts ...ClientOpts) (*OpkClient, error) {
+	client := &OpkClient{
+		Op:     op,
+		signer: nil,
+		alg:    nil,
+		signGQ: false,
+	}
+
+	for _, applyOpt := range opts {
+		applyOpt(client)
+	}
+
+	if client.alg == nil && client.signer != nil {
+		return nil, fmt.Errorf("signer specified but alg is nil, must specify alg  of signer")
+	}
+
+	if client.signer == nil {
+		// Generate signer for specified alg, if no alg specified choose ES256
+		if client.alg == nil {
+			client.alg = jwa.ES256
+		}
+
+		signer, err := util.GenKeyPair(client.alg)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create key pair for client: %w ", err)
+		}
+		client.signer = signer
+	}
+
+	return client, nil
+}
+
+type AuthOptsStruct struct {
+	extraClaims map[string]any
+}
+type AuthOpts func(a *AuthOptsStruct)
+
+// WithExtraClaims specifies additional values to be included in the
+// CIC. These claims will be include in the CIC protected header and
+// will be hashed into the commitment claim in the ID Token. The
+// commitment claim is typically the nonce or aud claim in the ID Token).
+// Example use:
+//
+//	WithExtraClaims(map[string]any{"claimKey": "claimValuez"})
+func WithExtraClaims(extraClaims map[string]any) AuthOpts {
+	return func(a *AuthOptsStruct) {
+		// Replace with maps.Clone when Clone is no longer experimental
+		a.extraClaims = map[string]any{}
+		for k, v := range extraClaims {
+			a.extraClaims[k] = v
+		}
+	}
 }
 
 // Auth returns a PK Token by running the OpenPubkey protocol. It will first
@@ -50,26 +138,17 @@ type OpkClient struct {
 //
 // signGQ specifies if the OPs signature on the ID Token should be replaced
 // with a GQ signature.
-func (o *OpkClient) Auth(ctx context.Context) (*pktoken.PKToken, error) {
-	if o.Alg == nil {
-		o.Alg = jwa.ES256
+func (o *OpkClient) Auth(ctx context.Context, opts ...AuthOpts) (*pktoken.PKToken, error) {
+	authOpts := &AuthOptsStruct{
+		extraClaims: map[string]any{},
 	}
-
-	if o.Signer == nil {
-		signer, err := util.GenKeyPair(o.Alg)
-		if err != nil {
-			return nil, fmt.Errorf("failed to create key pair for client: %d ", err)
-		}
-		o.Signer = signer
-	}
-
-	if o.ExtraClaims == nil {
-		o.ExtraClaims = map[string]any{}
+	for _, applyOpt := range opts {
+		applyOpt(authOpts)
 	}
 
 	// If no Cosigner is set then do standard OIDC authentication
 	if o.CosP == nil {
-		return o.OidcAuth(ctx, o.Signer, o.Alg, o.ExtraClaims, o.SignGQ)
+		return o.OidcAuth(ctx, o.signer, o.alg, authOpts.extraClaims, o.signGQ)
 	}
 
 	// If a Cosigner is set then check that will support doing Cosigner auth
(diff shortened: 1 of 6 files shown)


client/client.go
```diff
@@ -166,3 +262,15 @@ func (o *OpkClient) OidcAuth(
...
+	return o.signer
+}
+
+func (o *OpkClient) GetAlg() jwa.KeyAlgorithm {
+	return o.alg
+}
+
+func (o *OpkClient) GetSignGQ() bool {
+	return o.signGQ
+}
```
lgmugnier, line 276: What's the value of returning this variable?

client/client.go
```diff
@@ -31,11 +31,106 @@ import (
...
 	"github.com/openpubkey/openpubkey/pktoken"
 	"github.com/openpubkey/openpubkey/pktoken/clientinstance"
+	"github.com/openpubkey/openpubkey/util"
 )
 
 type OpkClient struct {
-	Op   OpenIdProvider
-	CosP *CosignerProvider
+	Op     OpenIdProvider
+	CosP   *CosignerProvider
```
lgmugnier, line 39: I think it makes the most sense to private these variables as well so that we can control creation in the `New` function

client/client.go
```diff
@@ -45,16 +140,17 @@ type OpkClient struct {
 //
 // signGQ specifies if the OPs signature on the ID Token should be replaced
 // with a GQ signature.
```
lgmugnier, line 142: I believe this can be removed now

client/client.go
```diff
@@ -31,11 +31,106 @@ import (
...
+
+type AuthOptsStruct struct {
+	extraClaims map[string]any
+}
+type AuthOpts func(a *AuthOptsStruct)
+
+// WithExtraClaims specifies additional values to be included in the
+// CIC. These claims will be include in the CIC protected header and
+// will be hashed into the commitment claim in the ID Token. The
+// commitment claim is typically the nonce or aud claim in the ID Token).
```
lgmugnier, line 122:
```suggestion
// commitment claim is typically the nonce or aud claim in the ID Token.
```

client/client.go
```diff
@@ -31,11 +31,106 @@ import (
...
+	for _, applyOpt := range opts {
+		applyOpt(client)
+	}
+
+	if client.alg == nil && client.signer != nil {
+		return nil, fmt.Errorf("signer specified but alg is nil, must specify alg  of signer")
+	}
+
+	if client.signer == nil {
+		// Generate signer for specified alg, if no alg specified choose ES256
```
lgmugnier, line 99:
```suggestion
		// Generate signer for specified alg. If no alg specified, defaults to ES256
```

client/client.go
```diff
@@ -31,11 +31,106 @@ import (
...
+
+// WithSigner allows the caller to inject their own signer and algorithm.
+// Use this option if to generate to bring your own user key pair. If this
+// option is not set the OpkClient constructor will automatically generate
+// a signer, i.e., key pair.
+// Example use:
+//
+//	signer, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+//	WithSigner(signer, jwa.ES256)
+func WithSigner(signer crypto.Signer, alg jwa.KeyAlgorithm) ClientOpts {
```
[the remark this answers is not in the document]
lgmugnier: All this being said, I think it's fine to put in what you have and then I can put up a PR quickly after to do the above if it's something we want to do.

lgmugnier (commented):
Overall, this is a fantastic improvement that I'm super excited to see

client/client.go
```diff
@@ -31,11 +31,106 @@ import (
...
-	CosP *CosignerProvider
+	Op     OpenIdProvider
+	CosP   *CosignerProvider
+	signer crypto.Signer
+	alg    jwa.KeyAlgorithm
+	signGQ bool // Default is false
+}
+
+// ClientOpts contains options for constructing an OpkClient
+type ClientOpts func(o *OpkClient)
```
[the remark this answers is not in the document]
EthanHeilman: In trying to decide what to name this I looked up how options are named in golang official crypto library to try to keep things consistent. 

As far as I can tell they use `Options` when the full variable name is just `Options` such as in [src/crypto/ed25519/ed25519.go](https://cs.opensource.google/go/go/+/master:src/crypto/ed25519/ed25519.go;l=273;bpv=1;bpt=1?q=options&ss=go%2Fgo)
`func VerifyWithOptions(publicKey PublicKey, message, sig []byte, opts *Options)`

but when the are specifying a prefix they use Opts rather than Option as in [src/crypto/crypto.go](https://cs.opensource.google/go/go/+/refs/tags/go1.21.6:src/crypto/crypto.go;l=198):
`Sign(rand io.Reader, digest []byte, opts SignerOpts) (signature []byte, err error`

I decided to stick with this pattern from go crypto.

client/client.go
EthanHeilman: Good question:

1. It is an important property of the client so it seems likely that a developer may wish to check it is set the way they expect it.

2. client_test.go needs it to test that the option is set correctly https://github.com/openpubkey/openpubkey/pull/93/files#diff-43e8423bcc9bb4d9fcd3ce8da7e0197e64d49376e957292eadba0316d57b000dR73

client/client.go
EthanHeilman: Fixed

client/client.go
EthanHeilman: Fixed

client/client.go
EthanHeilman: +1 

My plan to do this is as follows:
1. Make cosP private since nothing external depends on it. Keep Op public for now.
2. Create Getters for Op and CosP. This will set us up for making OpkClient an interface.
3. Get PRs into downstream projects to move them to client.New() and OpkClient.Auth().
4. Once they are moved over, make op private.

I can do 1 and 2 in this PR and then 4 in a separate PR that makes OpkClient.OidcAuth() private as well

client/client.go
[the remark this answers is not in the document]
EthanHeilman: You are talking about the eternal dream of a SignerAndAlg interface?

client/client.go
EthanHeilman: Just checked in code that gets us 1 and 2. Once this PR is merged, I can create the PRs to switch everyone over to new.Client and  Auth(). None of the unittests or examples depend on cosP or Op being public values setting us up for 3.

jonnystoten (commented):
This is great stuff! :heart:

client/client.go
```diff
@@ -31,30 +31,123 @@ import (
...
+		alg:    nil,
+		signGQ: false,
+	}
+
+	for _, applyOpt := range opts {
+		applyOpt(client)
+	}
+
+	if client.alg == nil && client.signer != nil {
+		return nil, fmt.Errorf("signer specified but alg is nil, must specify alg  of signer")
```
jonnystoten, line 95:
```suggestion
		return nil, fmt.Errorf("signer specified but alg is nil, must specify alg of signer")
```

jonnystoten (changes requested):
and a tiny typo

client/client.go
```diff
@@ -31,30 +31,123 @@ import (
...
+type AuthOpts func(a *AuthOptsStruct)
+
+// WithExtraClaims specifies additional values to be included in the
+// CIC. These claims will be include in the CIC protected header and
+// will be hashed into the commitment claim in the ID Token. The
+// commitment claim is typically the nonce or aud claim in the ID Token.
+// Example use:
+//
+//	WithExtraClaims(map[string]any{"claimKey": "claimValuez"})
+func WithExtraClaims(extraClaims map[string]any) AuthOpts {
```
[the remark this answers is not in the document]
EthanHeilman: This is great!

client/client.go
```diff
@@ -68,11 +161,11 @@ func (o *OpkClient) Auth(
 			http.Redirect(w, r, redirectUri, http.StatusFound)
 		})
 
-		pkt, err := o.OidcAuth(ctx, signer, alg, extraClaims, signGQ)
+		pkt, err := o.OidcAuth(ctx, o.signer, o.alg, authOpts.extraClaims, o.signGQ)
```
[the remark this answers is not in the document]
EthanHeilman: The plan is to keep `OidcAuth` and some other functions as is to minimize how much we break downstream. Once we have made all the changes we want to make, we will create PRs to move everyone (currently just docker and SSHpiper) to the using the constructor and the `Auth` function. Then with no one relying on the old functions, in one fell swoop make `OidcAuth` private and change its params, remove the verifier that doesn't live in verifier package, and no longer expose the Op in the client struct.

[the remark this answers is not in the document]
jonnystoten: Ah got you, yeah I see the comment explaining that `OidcAuth` is deprecated now.

client/client.go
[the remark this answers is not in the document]
EthanHeilman: Implemented your idea. Should be fixed

client/client.go
```diff
@@ -31,30 +31,123 @@ import (
...
+// Example use:
+//
+//	WithExtraClaims(map[string]any{"claimKey": "claimValuez"})
+func WithExtraClaims(extraClaims map[string]any) AuthOpts {
+	return func(a *AuthOptsStruct) {
+		// Replace with maps.Clone when Clone is no longer experimental
+		a.extraClaims = map[string]any{}
+		for k, v := range extraClaims {
+			a.extraClaims[k] = v
+		}
```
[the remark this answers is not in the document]
EthanHeilman: Fixed

jonnystoten (approved):
LGTM!

PR #95, opened 2024-02-01
Adds the approved openpubkey Google clientid to all examples

EthanHeilman:
This PR standardizes the examples apps to all use the same Google clientid. This clientid is approved by google and allows anyone to run the examples and the consent screen reads OpenPubkey. Some of clientid we were using had confusing consent screens.

Also deletes some unused configuration from the opkssh example.

I have manually tested this on all three examples.

### DIFF
--- a/examples/google/example.go
+++ b/examples/google/example.go
@@ -41,9 +41,9 @@ import (
 
 // Variables for building our google provider
 var (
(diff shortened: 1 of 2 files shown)


lgmugnier (approved):
I love to see removing unused variables!

PR #97, opened 2024-02-04
Moves ca package into examples

EthanHeilman:
The `ca` package is an example and not currently functionality that should be part of the core functionality. This PR turns the ca package into an example and removes some duplicated code between the `ca` and `cert` packages.

This adds the OIDC issuer extension to the x509 used by the cert package. @mrjoelkamp you build stuff with this code, is this an OK change? Any reason not to do this?

```
ExtraExtensions: []pkix.Extension{{
	// OID for OIDC Issuer extension
	Id:       asn1.ObjectIdentifier{1, 3, 6, 1, 4, 1, 57264, 1, 1},
	Critical: false,
	Value:    []byte(idtClaims.Issuer),
}},
```

## Notes

* Used [`git mv` ](https://git-scm.com/docs/git-mv) for `ca.go` and `ca_test.go` but too many changes so git decided it was a delete and create rather than a move. >=(


cert/cert.go
```diff
@@ -19,44 +37,65 @@ import (
...
 		NotBefore:               time.Now(),
 		NotAfter:                time.Now().Add(365 * 24 * time.Hour), // valid for 1 year
 		KeyUsage:                x509.KeyUsageDigitalSignature,
 		ExtKeyUsage:             []x509.ExtKeyUsage{x509.ExtKeyUsageCodeSigning},
 		BasicConstraintsValid:   true,
-		DNSNames:                []string{payload.Subject},
+		DNSNames:                []string{idtClaims.Subject},
 		IsCA:                    false,
-		SubjectKeyId:            pkTokenJSON,
+		ExtraExtensions: []pkix.Extension{{
```
EthanHeilman, line 91: I should probably make this optional and have it controllable via a bool arg.

mrjoelkamp: This is probably OK without making it optional since usage here will always be in the context of a PK Token that has been issued by a provider.

mrjoelkamp (approved):
Yes, LGTM! It's currently being used [here](https://github.com/openpubkey/signed-attestation/blob/5ebc1df5bc0719b1b12dd73849efb9c7f481e2ff/tl.go#L102) and I don't see any changes that will break it.

jonnystoten:
> * Used [`git mv` ](https://git-scm.com/docs/git-mv) for `ca.go` and `ca_test.go` but too many changes so git decided it was a delete and create rather than a move. >=(

That's annoying! I wouldn't worry for this PR but in future it can be better to put the move itself in its own commit and then make the changes in further commits.

jonnystoten (approved):
LGTM

EthanHeilman:
> That's annoying! I wouldn't worry for this PR but in future it can be better to put the move itself in its own commit and then make the changes in further commits.

Maybe I have to do a mv PR and then a change PR because it seems like the PR merge Squash hides the mv, but I haven't tried the mv being the first commit in the PR. I should run some experiments.

PR #106, opened 2024-02-27
Fixes bug in which JKT was being base64 encoded twice

EthanHeilman:
We base64 encode jkt (public key thumbprint) in the public header of the OP signature. However our json marshaller will automatically base64 encode an unknown header if it has the type []byte. 

This PR fixes this by changing the type of the jkt in the public header to string, rather than []byte.

This PR also adds a regression test for this bug fix.


kipz (approved):
LGTM. Thanks! :)

PR #110, opened 2024-03-06
Removes Compact from PK Token struct

EthanHeilman:
PR #105 deprecated the `pkt.Compact` function as it is not safe to use. This PR removes this `pkt.Compact` completely and all rewrites all internal uses.

## Impacted projects:

To ensure removing `Compact` does not impact other projects, I've made a best effort to see if it used outside of the OpenPubkey repo.

* A github search of the openpubkey org for "Compact" shows that compact is only being used by the openpubkey repo.
* A github search for " openpubkey finds a number of projects using openpubkey. None of these projects would be impacted
    * [Secure Openpubkey Shell (SOS)](https://github.com/devlup-labs/sos/) - uses OpenPubkey but all compact uses are just copies of the OpenPubkey internal code. No dependencies on pkt.Compact.
    * [ext-openpubkey](https://github.com/kipz/ext-openpubkey) is just a fork of OpenPubkey. No dependencies on pkt.Compact.
    * Our internal zli dependency uses pkt.Compact but is points to it's own version of OpenPubkey so there is no impact.
    * [sshpiper-openpubkey](https://github.com/tg123/sshpiper-openpubkey) does not use [Compact](https://github.com/search?q=repo%3Atg123%2Fsshpiper-openpubkey%20Compact&type=code).

There does not appear to be any impacts on downstream.


mrjoelkamp (approved):
LGTM

PR #131, opened 2024-03-15
Creates code coverage github action

EthanHeilman:
Adds the ncruces/go-coverage-report@v0 code coverage tool to our github actions workflow. This is only run after a PR is merged to main. It is not run on every pull request.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -5,10 +5,6 @@ on:
     paths:
       - '**.go'
 
-permissions:
-  contents: write
-  pages: write
-
 jobs:
   gofmt:
     name: Check Code Formatting
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -5,6 +5,10 @@ on:
     paths:
       - '**.go'
 
+permissions:
+  contents: write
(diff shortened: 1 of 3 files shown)


.github/workflows/go.yml
```diff
@@ -56,3 +60,5 @@ jobs:
 
     - name: Test
       run: go test ./...
+
+
```
mrjoelkamp, line 64:
```suggestion

```

.github/workflows/go.yml
```diff
@@ -5,6 +5,10 @@ on:
     paths:
       - '**.go'
 
+permissions:
+  contents: write
+  pages: write
+
```
mrjoelkamp, line 11: what are these permissions for? I see they are required for `ncruces/go-coverage-report` in the other workflow

EthanHeilman: Excellent catch! I created the second workflow specifically so that the go.yml wouldn't get these permissions.

EthanHeilman: Fixed!

mrjoelkamp (approved):

PR #139, opened 2024-03-19
Removes unused function argument in x509/example

EthanHeilman:
Three line change, removing an unused function argument in an example. Merging this gets us to zero golang warnings.


mrjoelkamp (approved):
🎉

mrjoelkamp (approved):

PR #141, opened 2024-03-20
Pass pointer to key.Raw()

ymarcus93:
When testing with Microsoft as the OP, I received the following error:

```
failed to decode public key: destination argument to AssignIfCompatible() must be a pointer: <nil>
```

from:

https://github.com/openpubkey/openpubkey/blob/83be2414ed515ea18390b519c3c14936a1e05671/client/providers/discover/discover.go#L36-L39

After further investigation, it seems that we're failing because in the case of Microsoft `key.Algorithm() == nil`, so we're attempting the default case where we don't know the exact type of `jwk.Key`. Following the documentation of `jwk.Key.Raw()`,

> If you do not know the exact type of a jwk.Key before attempting to obtain the raw key, you can simply pass a pointer to an empty interface as the first argument.

I've changed the invocation of `Raw()` to pass a pointer to `pubKey`.

### DIFF
--- a/client/providers/discover/discover.go
+++ b/client/providers/discover/discover.go
@@ -29,18 +29,31 @@ func NewPublicKeyRecord(key jwk.Key, issuer string) (*PublicKeyRecord, error) {
 	var pubKey interface{}
 	if key.Algorithm() == jwa.RS256 {
 		pubKey = new(rsa.PublicKey)
-	}
-	if key.Algorithm() == jwa.ES256 {
+	} else if key.Algorithm() == jwa.ES256 {
 		pubKey = new(ecdsa.PublicKey)
+	} else if key.Algorithm().String() == "" {
+		// OPs such as azure (microsoft) do not specify alg in their JWKS. To
+		// handle this case, assume no alg in JWKS means RSA as OIDC requires
+		// OPs use RSA.
+		pubKey = new(rsa.PublicKey)
+	} else {
+		return nil, fmt.Errorf("JWK has unsupported alg (%s)", key.Algorithm())
 	}
 	err := key.Raw(&pubKey)
 	if err != nil {
 		return nil, fmt.Errorf("failed to decode public key: %w", err)
 	}
 
+	var alg string
+	if key.Algorithm().String() == "" {
+		alg = jwa.RS256.String()
+	} else {
+		alg = key.Algorithm().String()
+	}
+
 	return &PublicKeyRecord{
 		PublicKey: pubKey,
-		Alg:       key.Algorithm().String(),
+		Alg:       alg,
 		Issuer:    issuer,
 	}, nil
 }
--- a/client/providers/discover/discover.go
+++ b/client/providers/discover/discover.go
@@ -33,7 +33,7 @@ func NewPublicKeyRecord(key jwk.Key, issuer string) (*PublicKeyRecord, error) {
 	if key.Algorithm() == jwa.ES256 {
 		pubKey = new(ecdsa.PublicKey)
 	}
-	err := key.Raw(pubKey)
+	err := key.Raw(&pubKey)
 	if err != nil {
 		return nil, fmt.Errorf("failed to decode public key: %w", err)
 	}
--- a/client/providers/discover/discover_test.go
+++ b/client/providers/discover/discover_test.go
@@ -36,6 +36,81 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestNewPublicKeyRecord(t *testing.T) {
+	commonIssuer := "https://example.com"
+	tests := []struct {
+		name        string
+		keyJson     map[string]string
+		expectedAlg string
+		shouldError bool
+	}{
+		{
+			name: "alg=RS256",
+			keyJson: map[string]string{
+				jwk.AlgorithmKey: "RS256",
+				jwk.KeyTypeKey:   "RSA",
+				jwk.RSAEKey:      "AQAB",
+				jwk.RSANKey:      "0vx7agoebGcQSuuPiLJXZptN9nndrQmbXEps2aiAFbWhM78LhWx4cbbfAAtVT86zwu1RK7aPFFxuhDR1L6tSoc_BJECPebWKRXjBZCiFV4n3oknjhMstn64tZ_2W-5JsGY4Hc5n9yBXArwl93lqt7_RN5w6Cf0h4QyQ5v-65YGjQR0_FDW2QvzqY368QQMicAtaSqzs8KJZgnYb9c7d0zgdAZHzu6qMQvRL5hajrn1n91CbOpbISD08qNLyrdkt-bFTWhAI4vMQFh6WeZu0fM4lFd2NcRwr3XPksINHaQ-G_xBniIqbw0Ls1jF44-csFCur-kEgU8awapJzKnqDKgw",
+			},
+			expectedAlg: jwa.RS256.String(),
+		},
+		{
+			name: "alg=ES256",
+			keyJson: map[string]string{
+				jwk.AlgorithmKey: "ES256",
+				jwk.KeyTypeKey:   "EC",
+				jwk.ECDSACrvKey:  "P-256",
+				jwk.ECDSAXKey:    "MKBCTNIcKUSDii11ySs3526iDZ8AiTo7Tu6KPAqv7D4",
+				jwk.ECDSAYKey:    "4Etl6SRW2YiLUrN5vfvVHuhp7x8PxltmWWlbbM4IFyM",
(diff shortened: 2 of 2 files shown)


EthanHeilman (changes requested):
Great catch and nice write up in the PR.

Before merging, can you add a unittest to discover_test.go so that can ensure we catch this bug going forward.

client/providers/discover/discover.go
```diff
@@ -33,7 +33,7 @@ func NewPublicKeyRecord(key jwk.Key, issuer string) (*PublicKeyRecord, error) {
 	if key.Algorithm() == jwa.ES256 {
 		pubKey = new(ecdsa.PublicKey)
 	}
-	err := key.Raw(pubKey)
+	err := key.Raw(&pubKey)
```
EthanHeilman, line 36:
```suggestion
    if key.Algorithm() ==nil {
        // OPs such as azure (microsoft) does not specify alg in their JWKS. To handle this case 
        // assume no alg in JWKS means RSA. as OIDC says OPs must use RSA 
        pubKey = new(rsa.PublicKey)
	}
    err := key.Raw(&pubKey)
```

EthanHeilman:
Additional details:

Microsoft does not specify the Alg of their keys are their JWKS URI: https://login.microsoftonline.com/common/discovery/v2.0/keys

```json
{
    "keys": [
        {
            "kty": "RSA",
            "use": "sig",
            "kid": "XRvko8P7A3UaWSnU7bM9nT0MjhA",
            "x5t": "XRvko8P7A3UaWSnU7bM9nT0MjhA",
            "n": "vRIL3aZt-xVqOZgMOr71ltWe9YY2Wf_B28C4Jl2nBSTEcFnf_eqOHZ8yzUBbLc4Nti2_ETcCsTUNuzS368BWkSgxc45JBH1wFSoWNFUSXaPt8mRwJYTF0H32iNhw_tBb9mvdQVgVs4Ci0dVJRYiz-ilk3PeO8wzlwRuwWIsaKFYlMyOKG9DVFbg93DmP5Tjq3C3oJlATyhAiJJc1T2trEP8960an33dDEaWwVAHh3c_34meAO4R6kLzIq0JnSsZMYB9O_6bMyIlzxmdZ8F442SynCUHxhnIh3yZew-xDdeHr6Ofl7KeVUcvSiZP9X44CaVJvknXQbBYNl-H7YF5RgQ",
            "e": "AQAB",
            "x5c": [
                "MIIC/jCCAeagAwIBAgIJAKysonliFZLIMA0GCSqGSIb3DQEBCwUAMC0xKzApBgNVBAMTImFjY291bnRzLmFjY2Vzc2NvbnRyb2wud2luZG93cy5uZXQwHhcNMjQwMjA4MTcwMjUzWhcNMjkwMjA4MTcwMjUzWjAtMSswKQYDVQQDEyJhY2NvdW50cy5hY2Nlc3Njb250cm9sLndpbmRvd3MubmV0MIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIBCgKCAQEAvRIL3aZt+xVqOZgMOr71ltWe9YY2Wf/B28C4Jl2nBSTEcFnf/eqOHZ8yzUBbLc4Nti2/ETcCsTUNuzS368BWkSgxc45JBH1wFSoWNFUSXaPt8mRwJYTF0H32iNhw/tBb9mvdQVgVs4Ci0dVJRYiz+ilk3PeO8wzlwRuwWIsaKFYlMyOKG9DVFbg93DmP5Tjq3C3oJlATyhAiJJc1T2trEP8960an33dDEaWwVAHh3c/34meAO4R6kLzIq0JnSsZMYB9O/6bMyIlzxmdZ8F442SynCUHxhnIh3yZew+xDdeHr6Ofl7KeVUcvSiZP9X44CaVJvknXQbBYNl+H7YF5RgQIDAQABoyEwHzAdBgNVHQ4EFgQU8Sqmrf0UFpZbGtl5y1CjUdQq5ycwDQYJKoZIhvcNAQELBQADggEBAA57FiIOUs5yyLD6a6rWCbQ4Z2XJTfQb+TM/tZ6V6QqNhSS+Q98KFOIWe9Sit0iAyDsCCKuA8f08PYnUiHmHq8dG/7YRTShE/3zCZXHYKJgMaBhYfS788zQuq/hXDdVVc5X0pZwM4ibc6+2XpcpeDHxpMOLwo2AwujDdHVLzedAkIaTCzwPIizP4LB6l6IxR+xRXsH/1f034Gk3ReAEGgHW12NkajtXmC3DKl6vGIHvx1PgAMWQbxq3F2OopNx6aZEIIZWcMpQZ6/62f3pxRJHzZiJZN+khV8hpNjJvCNf6/hNbxkLcjLAycjW8AttcCRSTM4F+02S3TyHmoE4pYywA="
            ],
            "issuer": "https://login.microsoftonline.com/{tenantid}/v2.0"
        },
        ...
}
```

But their .well-known/openid-configuration URI https://login.microsoftonline.com/common/.well-known/openid-configuration does specify that  `"id_token_signing_alg_values_supported": ["RS256"]`. That is all ID Tokens are signed with RSA, so all keys in the JWKS endpoint can be assumed to be RSA.

```json
{
    "token_endpoint": "https://login.microsoftonline.com/common/oauth2/token",
    "token_endpoint_auth_methods_supported": [
        "client_secret_post",
        "private_key_jwt",
        "client_secret_basic"
    ],
    "jwks_uri": "https://login.microsoftonline.com/common/discovery/keys",
    "response_modes_supported": [
        "query",
        "fragment",
        "form_post"
    ],
    "subject_types_supported": [
        "pairwise"
    ],
    "id_token_signing_alg_values_supported": [
        "RS256"
    ],
    "response_types_supported": [
        "code",
        "id_token",
        "code id_token",
        "token id_token",
        "token"
    ],
    "scopes_supported": [
        "openid"
    ],
    "issuer": "https://sts.windows.net/{tenantid}/",
    "microsoft_multi_refresh_token": true,
    "authorization_endpoint": "https://login.microsoftonline.com/common/oauth2/authorize",
    "device_authorization_endpoint": "https://login.microsoftonline.com/common/oauth2/devicecode",
    "http_logout_supported": true,
    "frontchannel_logout_supported": true,
    "end_session_endpoint": "https://login.microsoftonline.com/common/oauth2/logout",
    "claims_supported": [
        "sub",
        "iss",
        "cloud_instance_name",
        "cloud_instance_host_name",
        "cloud_graph_host_name",
        "msgraph_host",
        "aud",
        "exp",
        "iat",
        "auth_time",
        "acr",
        "amr",
        "nonce",
        "email",
        "given_name",
        "family_name",
        "nickname"
    ],
    "check_session_iframe": "https://login.microsoftonline.com/common/oauth2/checksession",
    "userinfo_endpoint": "https://login.microsoftonline.com/common/openid/userinfo",
    "kerberos_endpoint": "https://login.microsoftonline.com/common/kerberos",
    "tenant_region_scope": null,
    "cloud_instance_name": "microsoftonline.com",
    "cloud_graph_host_name": "graph.windows.net",
    "msgraph_host": "graph.microsoft.com",
    "rbac_url": "https://pas.windows.net"
}
```

client/providers/discover/discover.go
ymarcus93: Pushed a0bfa92 and added unittest to `discover_test.go`. However, is it fair to assume if alg is not specified this means RSA and `RS256`? What if it means RSA but sig alg is `RS384` or `RS512`? OIDC spec says OP must use RSA but how do we know if missing alg actually means the provided key is using RSA, or are we just assuming that?

ymarcus93: Ok. After further examination in the OIDC spec, it does say that `RS256` is the default, so this seems fine to assume the alg as well.

> The alg value SHOULD be the default of RS256 or the algorithm sent by the Client in the id_token_signed_response_alg parameter during Registration.

EthanHeilman: The OIDC spec says:
> OPs MUST support signing ID Tokens with the RSA SHA-256 algorithm (an alg value of RS256), unless the OP only supports returning ID Tokens from the Token Endpoint (as is the case for the Authorization Code Flow) and only allows Clients to register specifying none as the requested ID Token signing algorithm. 

https://openid.net/specs/openid-connect-core-1_0.html#ServerMTI

So RS256 is the requirement not just RSA. That said, I like that you are thinking about this and it is something we should consider, not all OPs follow the spec [and other projects have run into issues with OPs signing ID Tokens with RS512.](https://community.auth0.com/t/getting-rs512-error-when-using-openid-connect/68579)

EthanHeilman (approved):
LGTM

client/providers/discover/discover.go
EthanHeilman: Jinks!

Issue #26, opened 2023-10-06
Add GitLab CI signing example


aarongoldenthal-work:
Given the broad use of GitLab in environments where an OIDC provider as a CA has significant benefits, it would be helpful to have a reference example.

EthanHeilman:
@aarongoldenthal-work Thanks, it would be great to support gitlab. I am looking into it, but will probably be a little while before this is added.

We are accepting PRs if you wanted to write a proof-of-concept I'd be happy to work with you to get merged.

kipz:
It looks as if GitLab OIDC provider for workloads/automation doesn't support setting any claims on-the-fly, so we may not be able to bind the ephemeral keys to the identity. 

@EthanHeilman @jonnystoten should we consider a mode for workload/code-signing whereby we use GQ to sign directly and don't generate and use the ephemeral keys at all?

IMHO, having a story for GitLab is super important.

jonnystoten:
I'm interested in the idea of using GQ to sign the CIC. I think this means we don't need to bind the key using the nonce/audience claim. This could be useful for GitHub as well, because it would allow us to create multiple PK tokens with different CICs from the same ID token. The disadvantage is that the original OIDC token is more valuable if compromised. Personally I think that's a reasonable tradeoff, keen to hear other opinions though.

I can create another issue for GQ signing the CIC so we can keep this one focused on GitLab integration.

EthanHeilman:
>  The disadvantage is that the original OIDC token is more valuable if compromised. Personally I think that's a reasonable tradeoff, keen to hear other opinions though.
@jonnystoten I think it is a reasonable trade off as well. However in this setting you need to be very careful that you don't use the audience claim for non-GQ.

I was playing around with the right way to represent this last night:

Set the audience claim to  `aud=GQ-Only-<Intended Audience>`. If we see a token with this audience claim prefix we should ensure it is from the list of GQ only issuers such as gitlab and reject otherwise. I believe this can be fit in with the proposed GQ JWS representation mechanism from: https://github.com/openpubkey/openpubkey/issues/15#issuecomment-1730255388

ID Token you get from Gitlab
```golang
{
    "payload":  BASE64URL({
    "aud": GQ-Only-<Intended Audience>,
    ...
    }), // ID Token
    "signatures": [
        {
            "protected": BASE64URL({"alg":"RSA256","kid":"6","typ":"JWT"}),
            "signature": BASE64URL(RSAsig) // OpenID Provider Signature
        },
```
[...]

Issue #100, opened 2024-02-19
Design for Gitlab CI support 


EthanHeilman:
As discussed in [Issue 26: Add GitLab CI signing example](https://github.com/openpubkey/openpubkey/issues/26) the approach used to sign software artifacts in github can not be used with gitlab because the gitlab workload can not specify the aud (audience) claim. Instead an aud claim is hardcoded in the workload configuration YAML. In this issue we propose the following design for gitlab and other settings where there is no client specified nonce or audience field.

## Generating a PK Token in Gitlab

To generate a PK Token:
1. The workload creates a key pair, (upk, usk), and requests an ID Token from the OP (gitlab).
2. On receiving the ID Token from gitlab, the workload: generates the CIC (upk, alg, nonce), signs the CIC and ID token under the GQ signature.
3. The workload then immediate deletes the OPs RSA signature on the ID Token.

The CIC is included as part of the GQ protected header of the ID Token. Thus, when the GQ signature signs the new protected header, it also commits to the CIC. As is currently done with GQ signatures, the original protected header of the ID Token is base64 encoded and included in the protected header signed by the GQ signature.

```
payload: {
  "iss": "gitlab.example.com",
  "aud": "OPENPUBKEY-PKTOKEN:1234567890",
  ...
} 
signatures: [
  {"protected": {"typ": "JWT", "alg": "GQ256", "kid":  Base64(origProtected), 
  "cic": SHA3_256(upk=alice-pubkey, alg=ES256, rz=crypto.Rand()),
  "signature": GQSIGN(RSA-sig, (payload, signatures[0].protected))`
  },
  {"protected": {"upk": alice-pubkey, "alg": "EC256", "rz": <rz>},
  "signature": <SIGN(alice-signkey, (payload, signatures[1].protected))>
  },
]
```

## Verifying a Gitlab PK Token

1. Check the aud claim is prefixed with "OPENPUBKEY-PKTOKEN-" and that the rest of the aud claim matches the expected aud claim.
2. Check that CIC signature verifies,
3. Check that the CIC protected header values are committed to in the cic claim in protected header of GQ signature and that the GQ signature verifies for gitlabs public key.

## Security

If configured correctly, this is approach offers the same security as the approach used with github. However this approach does introduce a  security risk of misconfiguration not present in the github case. If a GQ signature is not used and an attacker learns the ID Token, the attacker can compute the GQ signature from the RSA signature and insert their own public key. 

We mitigate the risk of this occurring by:

[...]

PR #143, opened 2024-03-20
Adds gitlab-ci OP using GQ commitment binding

EthanHeilman:
This PR adds a gitlab-ci OP to OpenPubkey and assorted tests. Testing this feature required adding a number of new testing features and building out ProviderVerifier table based unittests.

## GQ Bound PK Tokens

For google and github we bind the CIC (Client Instance Claims) that contains the user's public key to the ID Token by putting a hashed commitment in one of the claims of the ID Token. For Google we put the commitment in the nonce claim, for github we put the commitment in the audience (aud) claim. Gitlab-CI does not provide an nonce claim and does not allow a running job/workflow to specify the aud. For gitlab the value of aud claim can only be specified in the static workflow config. This presents a problem: how to bind the CIC to the ID Token.

We provide a solution that solves this problem for gitlab-ci but also solves this problem more generally. We use the OP's RSA signature on the ID Token to create a GQ signature and then we delete the RSA signature. This GQ signature, functions as proof of knowledge of the RSA signature, enabling third parties to check that the ID Token was really signed by the OP. Additionally as the name suggests a GQ Signature is also a signature and use this signing property to bind the CIC to the ID Token. 

More specifically, put the commitment, a.k.a., the hash of the CIC, in the protected header of the GQ signed ID Token and then we GQ sign the payload and the protected header using the header claim "cic". Note that to ensure we don't lose the original protected header of the ID Token, we base64 encode it and set it as the kid (KeyID) in the GQ Signed ID Token's protected header. Below we show an example PK Token that uses a GQ commitment binding:

```go
payload: {
  "iss": "https://token.actions.gitlab.com",
  "aud": "OPENPUBKEY-PKTOKEN:1234",
  ...
} 
signatures: [
  {"protected": {"typ": "JWT", "alg": "GQ256", "kid":  Base64(origProtected), 
  "cic": SHA3_256(upk=alice-pubkey, alg=ES256, rz=crypto.Rand()),
  "signature": GQSIGN(RSA-sig, (payload, signatures[0].protected))`
  },
  {"protected": {"upk": alice-pubkey, "alg": "EC256", "rz": <rz>},
  "signature": <SIGN(alice-signkey, (payload, signatures[1].protected))>
  },
]
```

The design is based on the design proposed in https://github.com/openpubkey/openpubkey/issues/100 and should fix https://github.com/openpubkey/openpubkey/issues/26

## Security

If configured correctly, this is approach offers the same security as the approach used with github. However GQ bound PK Tokens do introduce a security risk from misconfiguration not present in the github case. The risk is that if a GQ signature is not used and an attacker learns the ID Token, the attacker can compute the GQ signature from the RSA signature and insert their own public key.

We mitigate the risk by:

* Requiring that PK Tokens that use the GQ signature to bind the CIC to the ID TOken must have "OPENPUBKEY-PKTOKEN:" prefixed in the audience and that they are not considered valid PK Tokens otherwise. This also prevents attacks in which a gitlab ID Token not intended for use as a PK Token is replayed as a PK Token. Note that this rule does not apply to other types of PK Tokens.
* Ensuring that OpenPubkey clients always enforces the deletion of the RSA signature and only generates GQ signature PK Tokens for gitlab. We use memguard to reduce the risk of exposure through process memory.

We argue that this sort of misconfiguration is very unlikely because outside of OpenPubkey, ID Tokens are treated as highly sensitive authentication secrets and within OpenPubkey we specifically prevent this misconfiguration from occurring. Someone must configure their gitlab to put "OPENPUBKEY-PKTOKEN:" in the aud claim, not use the OpenPubkey protocol and then in addition to that they must expose the ID token to an attacker. We can not prevent people from writing their own insecure protocols.

## TODOs
- [X] Add Gitlab OP
- [X] Add support for GQ commitment binding to gq library
- [X] Adversarial tests
- [X] Gitlab OP unittests
- [X] GQ commitment binding VerifierProvider unittests
- [X] Gitlab example
- [X] Create manual test to check this works on gitlab. Repo set for this purpose here: https://gitlab.com/openpubkey/gl-test
- [X] Allow specifying gitlab env variable to read ID Token from
- [X] Full PR description of change

## Tests

Successful OpenPubkey gitlab-CI job 
image

* Gitlab-CI: Manually tested this PR for the commit e1c3d804b65fb88d52fe756530e32ea3bd0141be https://gitlab.com/openpubkey/gl-test/-/jobs/6474201669

* Github Actions: Manually tested this PR for the commit e1c3d804b65fb88d52fe756530e32ea3bd0141be https://github.com/openpubkey/gha-test/actions/runs/8426700367/job/23075582010

*  I have successfully manually tested mfa cosigner on firefox/chrome/edge. MFA cosigner on brave is breaking with the error `Error during registration: NotAllowedError: The operation either timed out or was not allowed. See: https://www.w3.org/TR/webauthn-2/#sctn-privacy-considerations-client.`It error is not related to these changes in this PR because I also tested against main and got the same error. I suspect something in a recent update to brave broke this.

### DIFF
--- a/examples/gitlab/example.go
+++ b/examples/gitlab/example.go
@@ -1,3 +1,19 @@
+// Copyright 2024 OpenPubkey
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
 package gitlab_example
 
 import (
--- a/examples/gitlab/example.go
+++ b/examples/gitlab/example.go
@@ -27,42 +27,56 @@ import (
 
 func SignWithGitlab() error {
 
-	op, err := providers.NewGitlabOpFromEnvironment("OPENPUBKEY_JWT")
-	if err != nil {
-		return err
-	}
+	// Creates OpenID Provider (OP) configuration, this will be used to request the ID Token from Gitlab
+	op := providers.NewGitlabOpFromEnvironment("OPENPUBKEY_JWT")
+
+	// Creates a new OpenPubkey client
 	opkClient, err := client.New(op)
 	if err != nil {
 		return err
 	}
 
+	// Generates a PK Token by authorizing to the OpenID Provider
 	pkt, err := opkClient.Auth(context.Background())
 	if err != nil {
 		return err
 	}
 
+	// Serialize the PK Token to JSON so we can print it. Typically this
+	// serialization of the PK Token would be sent with the signed message
 	pktJson, err := pkt.MarshalJSON()
 	if err != nil {
 		return err
 	}
 	fmt.Println("pkt:", pktJson)
 
+	// Create a verifier to check that the PK Token is well formed
+	// The OPK client does this as well, but for the purposes of the
+	// example we show how a relying party might verify a PK Token
 	verifier, err := verifier.New(op.Verifier())
 	if err != nil {
 		return err
 	}
 
+	// Verify the PK Token. We supply the OP (gitlab) we wish to verify against
 	err = verifier.VerifyPKToken(context.Background(), pkt)
 	if err != nil {
 		return err
 	}
 
+	// Sign a message over the user's public key in the PK Token
 	msg := []byte("All is discovered - flee at once")
 	signedMsg, err := pkt.NewSignedMessage(msg, opkClient.GetSigner())
 	if err != nil {
 		return err
 	}
 	fmt.Println("signedMsg:", string(signedMsg))
+
+	// Verify the signed message
+	_, err = pkt.VerifySignedMessage(signedMsg)
+	if err != nil {
+		return err
+	}
 
 	fmt.Println("Success!")
 	return nil
--- a/examples/gitlab/example.go
+++ b/examples/gitlab/example.go
@@ -1,0 +1,53 @@
+package gitlab_example
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/openpubkey/openpubkey/client"
+	"github.com/openpubkey/openpubkey/client/providers"
+	"github.com/openpubkey/openpubkey/verifier"
+)
+
+func SignWithGitlab() error {
+
+	op, err := providers.NewGitlabOpFromEnvironment("OPENPUBKEY_JWT")
+	if err != nil {
+		return err
+	}
+	opkClient, err := client.New(op)
+	if err != nil {
+		return err
+	}
+
+	pkt, err := opkClient.Auth(context.Background())
+	if err != nil {
+		return err
+	}
+
+	pktJson, err := pkt.MarshalJSON()
+	if err != nil {
+		return err
+	}
+	fmt.Println("pkt:", pktJson)
+
+	verifier, err := verifier.New(op.Verifier())
+	if err != nil {
+		return err
+	}
+
+	err = verifier.VerifyPKToken(context.Background(), pkt)
+	if err != nil {
+		return err
+	}
+
+	msg := []byte("All is discovered - flee at once")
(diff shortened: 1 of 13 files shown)


examples/gitlab/example.go
```diff
@@ -0,0 +1,69 @@
...
+		return err
+	}
+
+	msg := []byte("All is discovered - flee at once")
+	signedMsg, err := pkt.NewSignedMessage(msg, opkClient.GetSigner())
+	if err != nil {
+		return err
+	}
+	fmt.Println("signedMsg:", string(signedMsg))
+
```
mrjoelkamp, line 66: Add `pkt.VerifySignedMessage()` for good measure?

examples/gitlab/example.go
```diff
@@ -0,0 +1,69 @@
...
+	op, err := providers.NewGitlabOpFromEnvironment("OPENPUBKEY_JWT")
+	if err != nil {
+		return err
+	}
+	opkClient, err := client.New(op)
+	if err != nil {
+		return err
+	}
+
```
mrjoelkamp, line 38: for examples, its always nice to have lots of comments describing each step

verifier/providerverifier.go
```diff
@@ -241,3 +300,13 @@ func verifyAudience(pkt *pktoken.PKToken, clientID string) error {
 	}
 	return nil
 }
+
+func VerifyCicSignature(pkt *pktoken.PKToken) error {
```
mrjoelkamp, line 304:
```suggestion
func verifyCicSignature(pkt *pktoken.PKToken) error {
```

Should this be private since it is important that the token's signature has been verified before the CIC is verified? So it should only ever be called from `VerifyProvider` after the pkt signature is verified?

mrjoelkamp (approved):

aarongoldenthal-work:
@EthanHeilman Thanks for letting me know. Overall it addresses what I was thinking in #26, I just saw three little things:

- It looks like [here](https://github.com/openpubkey/openpubkey/blob/bbfb1dc5b9731a1c3e98f9b480fdbba51a6b26b5/client/providers/gitlab_ci.go#L44) it's hardcoded to gitlab.com. Did I miss a place that could be specified? Is there a reason this couldn't also be supported with self-hosted installations as long as the host was specified?
- It looks like the code and tests are consistent in requiring the ID Token `aud` prefixed with `OPENPUBKEY-PKTOKEN:`, but the description above uses `OPENPUBKEY-PKTOKEN-` (in case anyone comes back here for reference later).
- Does the test repo https://gitlab.com/openpubkey/gl-test still exist, and is it possible it's private? I get a 404 when trying to open, which GitLab unfortunately gives for non-existent or no permissions so not sure, but I think a reference to a GitLab CI job with the appropriate settings (e.g. `id_token`) would be helpful.

examples/gitlab/example.go
EthanHeilman: Good point, added some comments

examples/gitlab/example.go
EthanHeilman: Fixed!

EthanHeilman:
@aarongoldenthal-work 

> It looks like [here](https://github.com/openpubkey/openpubkey/blob/bbfb1dc5b9731a1c3e98f9b480fdbba51a6b26b5/client/providers/gitlab_ci.go#L44) it's hardcoded to gitlab.com. Did I miss a place that could be specified? Is there a reason this couldn't also be supported with self-hosted installations as long as the host was specified?

I had not thought about that, but that is an excellent point. Created an additional constructor that allows anyone to specify the issuer for GitlabOP

```go
func NewGitlabOp(issuer string, tokenEnvVar string) *GitlabOp {
	op := &GitlabOp{
		issuer:          issuer,
		publicKeyFinder: *discover.DefaultPubkeyFinder(),
		getTokensFunc:   getEnvVar,
		tokenEnvVar:     tokenEnvVar,
	}
	return op
}
```

> It looks like the code and tests are consistent in requiring the ID Token aud prefixed with OPENPUBKEY-PKTOKEN:, but the description above uses OPENPUBKEY-PKTOKEN- (in case anyone comes back here for reference later).

Excellent catch, I'll fix that now. I changed that value so many times. Should be fixed now

> Does the test repo https://gitlab.com/openpubkey/gl-test still exist, and is it possible it's private? 

I thought I had made it public, but it was private. It is public now

https://gitlab.com/openpubkey/gl-test/

Thanks for all your comments. Let me know if you have anymore. If you end up using gitlab on OpenPubkey please open issuers for any problems or hit me up on email or slack.

PR #156, opened 2024-03-26
Provider and ID Token test overhaul

EthanHeilman:
This PR is a large redesign of OpenPubkey's testing infrastructure. It has three main goals:
1. Make testing of OpenID Providers possible. This is motivated by the bug that was not-caught by tests in the github-actions provider.
2. Increase code coverage, especially code coverage of failure conditions. This PR enables the ability to precisely inject badness into ID Tokens.
3. Break the provider, PK Token dependency. Right now the providers package imports the pktoken package, but the pktoken package tests depend on the provider package. This complicates the code base because one has to be very careful not to introduce circular dependencies. While this PR does not solves this problem, it lays the ground work for the next PR which will solve this issue. See PR here https://github.com/openpubkey/openpubkey/pull/163

## Mock Provider Backend 

This PR introduces the notion of a mock provider backend. Within this PR, this provides a place to inject a mock into providers, allowing much more complete tests of the providers.

 Long term this is intended to ease adding additional OPs (OpenId Providers) to OpenPubkey by providing a set of common utilities that providers can use to set provider public keys and to request tokens from that providers remote OP. This common set of utilities is not created in this PR, but the function hooks used by the MockProviderBackend provide a foundation to build this functionality later.

This PR enables us to unittest providers/google

### PKToken MockBackend circular dependency

This pr uses the package providers.backend rather than providers.mocks for the MockProviderBackend to avoid a circular dependency between the package pktoken and the package providers.mocks. This is a temporary fix to keep this PR focused on tests, and allow us to stage the PR that solves this separately.

## New MockProvider

The MockProvider constructor `NewMockProvider` returns the mock provider, the mockBackend for that provider and the ID Token Template for that provider.

```go
providerOpts := DefaultMockProviderOpts()
mockProvider, mockBackend, idtTemplate, err := NewMockProvider(providerOpts)
```

### ProviderOpts

ProviderOpts are options that allow you to specify how the provider should behave. It includes both options for the provider itself and options for the provider verifier. This lets a test create a provider which will issue tokens the verifier will reject to test the verifer. For instance we get the commit type of the provider to nonce and the commit type of the verifier to audience (aud). This lets us easily create test ID Tokens which and then test the verifier correctly catches that the PK Token is bad.

```go
MockProviderOpts{
		Issuer:     "https://accounts.example.com",
		ClientID:   clientID,
		SignGQ:     false,
		CommitType: providers.CommitTypesEnum.NONCE_CLAIM,
		VerifierOpts: providers.ProviderVerifierOpts{
			CommitType:        providers.CommitTypesEnum.AUD_CLAIM,
			ClientID:          clientID,
			SkipClientIDCheck: false,
			GQOnly:            false,
		},
	}
```

### IDToken Templates

In addition to this, to provide fine grained control over the ID Token's returned by the provider, we introduce ID Token templates. These allow the caller to specify what the ID Token returned should look like:

```go
IDTokenTemplate{
		CommitFunc: backend.AddAudCommit,
		Issuer:         "mockIssuer",
		Nonce:          "empty",
		NoNonce:        false,
		Aud:            "empty",
		KeyID:          "mockKeyID",
		NoKeyID:        false,
		Alg:            "RS256",
		NoAlg:          false,
	}
```
These can then be injected into the mockBackend so that the provider will return an ID Token with expected properties.

## TODOs
- [X] Replace type PublicKeyFunc with override in all provider unittests
- [X] Get tests running on all provider unittests
- [X] More test coverage of the examples
- [X] Replace Mock PK Token with new Mocks
- [X] Replace Mock Provider with new Mocks
- [X] Better name for override
- [X] Create constructors for mock Provider to reduce the amount of code needed to stand one up
- [X] Default Mock Provider Options to reduce boilerplate

## Tests
- [X]  Manual test of MFA Cosigner example  0e3318eea7593a4b58ef35f95f7c96173cb2121b
- [X]  Manual test of Gitlab-CI - https://gitlab.com/openpubkey/gl-test/-/jobs/6523206421
- [X]  Manual test of Github - https://github.com/openpubkey/gha-test/actions/runs/8513833980/job/23318311044
- [X] Manual test of SSH example against commit f46f8ea321232aa1e0dda90525c735686e1b6145


providers/backend/idtoken.go
```diff
@@ -0,0 +1,123 @@
...
+type CommitmentType struct {
+	ClaimCommitment bool
+	ClaimName       string
+}
+
+// TODO: Not sure if I should make IDTokenTemplate a generator
+type IDTokenTemplate2 interface {
+	AddCommit(cicHash string)
+	IssueToken() ([]byte, error)
+}
```
EthanHeilman, line 37: Should be deleted

```suggestion
```

providers/backend/idtoken.go
```diff
@@ -0,0 +1,123 @@
...
+
+// AddCommit adds the commitment to the CIC to the ID Token. The
+// CommitmentFunc is specifed allowing custom commitment functions to be specified
+func (t *IDTokenTemplate) AddCommit(cicHash string) {
+	t.CommitFunc(t, cicHash)
+}
+
+// TODO: rename t as it is confusing with t being used in tests
+
```
EthanHeilman, line 62: Delete this
```suggestion
```

providers/backend/backend_test.go
```diff
@@ -0,0 +1,126 @@
...
+		for _, tc := range testCases {
+			// skSet - set of signingKeys, recordSet - set of publicKeyRecords
+			skSet, recordSet, err := CreateKeySet(tc.issuer, tc.alg, numKeys)
+			if tc.expError != "" {
+				require.ErrorContains(t, err, tc.expError, tc.name)
+			} else {
+				require.NoError(t, err, tc.name)
+			}
+			CheckKeySets(t, tc.name, tc.issuer, tc.alg, skSet, recordSet)
+		}
```
mrjoelkamp, line 111: Sometimes its helpful to call the test cases as subtests `tc.Run()` for error handling in the test results

```suggestion
	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			// skSet - set of signingKeys, recordSet - set of publicKeyRecords
			skSet, recordSet, err := CreateKeySet(tc.issuer, tc.alg, numKeys)
			if tc.expError != "" {
				require.ErrorContains(t, err, tc.expError, tc.name)
			} else {
				require.NoError(t, err, tc.name)
			}
			CheckKeySets(t, tc.name, tc.issuer, tc.alg, skSet, recordSet)
		})
	}
```

mrjoelkamp (approved):
Nice, looks clean and easy to use, thanks!

examples/gitlab/example.go
```diff
@@ -25,10 +25,21 @@ import (
...
-	// Creates OpenID Provider (OP) configuration, this will be used to request the ID Token from Gitlab
-	op := providers.NewGitlabOpFromEnvironment("OPENPUBKEY_JWT")
+	// If an alternative OP is provided, use that instead of the default.
+	// Currently only used for testing where a mockOP is provided.
+	if len(opts) > 0 && opts[0].altOp != nil {
+		op = opts[0].altOp
+	} else {
+		// Creates OpenID Provider (OP) configuration, this will be used to request the ID Token from Gitlab
+		op = providers.NewGitlabOpFromEnvironment("OPENPUBKEY_JWT")
+	}
```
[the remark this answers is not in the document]
EthanHeilman: It could be, but probably the best way to do that would be to create a new provider.

PR #163, opened 2024-04-01
Isolates pktoken and provider packages

### LINKED ISSUE #158: Move OIDC helper functions into their own package
`client/oidc.go` should it's own package. Many of the other packages in OpenPubkey could make use of it, but since it is in client, it creates circular dependencies is imported by a package which client also imports.

This oidc package should be combined with simplejws which lives in `pktoken/simplejws/simplejws.go`

### LINKED ISSUE #160: VerifyProvider should take a ID Token rather than a PK Token
Providers and VerifierProviders currently have a function: VerifyProvider `VerifyProvider(ctx context.Context, pkt *pktoken.PKToken) error` which verifies a PK Token. I'm not sure this is the right approach because it means VerifyProviders and Providers must understand and depend on PK Tokens.

I would argue that VerifyProviders should only know about clientinstance.claims and ID Tokens. I propose the following two changes:

1. `RequestTokens(ctx context.Context, cic *clientinstance.Claims) ([]byte, error)` takes clientinstance.claims in the form of a simple map. This decouples providers from the PK Token package.
2. `VerifyProvider(ctx context.Context, pkt *pktoken.PKToken) error` should be changed to `VerifyProvider(ctx context.Context, idt []byte, cic [string]map[string]) error`

### LINKED ISSUE #162: After solving PK Token Circular Depenencies remove duplicate testing helper function GenCIC
There are two copies of genCIC. One lives in providers/test.go and the other in providers/mock/mockprovider.go. Deduplicate these once circular dependency issue is fixed from https://github.com/openpubkey/openpubkey/issues/160

EthanHeilman:
The purpose of this PR is to setup a stable long term package structure for OpenPubkey that minimizes dependencies between the packages to enable features and new providers to be added without requiring module-wide changes.

This PR makes the following changes:
* Providers no longer PK Tokens, but ID Tokens. ProviderVerifiers no longer verify PK Tokens but ID Tokens and CICs. This breaks the dependency between the providers package and the pktoken package. Verification of the CIC Token in the PK Token now happens in the verifier package. 
* Creates a new package oidc from simplejws and adds a ID Token parsing support.  Previously much of this functionality was inside of PK Token.
* Deletes unused or unneeded functions like ExtractClaim, oidp_server.go and errors.go.

## Bug Fixes in this PR

- Fixes #162 by solving circular dependences with better package structure
- Fixes #160 by changing provider and providerVerifier to use ID Token rather than PK Tokens
- Fixes #158 by creating oidc package

### What I like about this PR

We can make changes to the CIC and how the user signs the CIC without having those change touch the providers.

Cleans up the provider pktoken relationship, creates oidc package to make process on having us rely less on jwx.

### What I don't like about this PR

Touches so many files. Hopefully after this PR, the better package structure will reduce the number of files we have to touch.

The mockprovider doesn't live in providers/mocks. This is because the MockProvider depends on the ProviderVerifier which lives in provider, but all the tests in provider depend on the provider/mock package. Eventually we need an ID Token mock and PK Token mock, so that mock provider can be moved out of providers package placed in the providers/mock.

## Current Package Graph

Package Graph at Main

## After this PR

Package Graph at PR 163

## Tests

- [X] MFA-Cosigner Test 97998d786c7f613087e6c0cce6176e0b951a805e
- [X] Gitlab test https://gitlab.com/openpubkey/gl-test/-/pipelines/1239308324
- [X] Github test https://github.com/openpubkey/gha-test/actions/runs/8546255415/job/23416266634

### Graph visualization
Graphs for package relationships excluded the packages examples, utils, cert, pktoken.clientinstance and cosigner to cut down the noise. The command used is:
`goda graph -cluster -short ./... - github.com/openpubkey/openpubkey/examples/... github.com/openpubkey/openpubkey/cert/... github.com/openpubkey/openpubkey/cosigner/... github.com/openpubkey/openpubkey/util/...  github.com/openpubkey/openpubkey/pktoken/clientinstance/... | dot -Tsvg -o graph.svg`


client/oidc.go
mrjoelkamp, line 1: At some point I am guessing this will be removed completely?

mrjoelkamp (approved):

client/oidc.go
EthanHeilman: 100%, I was very tempted to remove this in this PR, but this PR already does too much. I've been keeping that function for compatibility reasons. Originally I had planned to remove it after the next release, but we've changed so much as this point, I don't see a good reason not to just remove it before the next release.

PR #167, opened 2024-04-04
Google OP can use different port is that port is already taken

### LINKED ISSUE #135: Allow specifying multiple redirectURI ports in GoogleOP
Currently the developer must choose a single redirectURI port when creating a GoogleOP to pass to the client. Unfortunately if the port of the redirectURI is currently in use, then the client will throw an error. Instead we allow multiple ports to be chosen so that we can fall over to another port is the first port is in use.

EthanHeilman:
Our Google OP (OpenID Provider) listens on a port on localhost to receive the redirect callback from the Google (Google sends the auth_code via this redirect). There is no guarantee that the port we choose to listen on is not already be in use by another program.

This PR fixes this by allowing multiple redirectURIs to specified with different ports. It will try the different ports until it runs out of ports to turn or until it finds an available localhost port.

This PR also cleans up how the Google configuration and makes it easier for someone to setup a Google OP.

Fixes  #135 

## Manual Tests

- [X] MFA Cosigner test
- [X] Simple Google test

Not running github or gitlab manual tests as this code does not impact either of those providers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,14 +24,9 @@ Let's walk through a simple message signing example. For conciseness we omit the
 We start by configuring the OP (OpenID Provider) our client and verifier will use. In this example we use Google as our OP.
 
 ```golang
-op := &providers.GoogleOp{
-  ClientID:     "992028499768-ce9juclb3vvckh23r83fjkmvf1lvjq18.apps.googleusercontent.com",
-  ClientSecret: "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", // The client secret is a public value
-  Scopes:       []string{"openid profile email"},
-  RedirURIPort: "3000",
-  CallbackPath: "/login-callback",
-  RedirectURI:  "http://localhost:3000/login-callback",
-}
+	opOptions := providers.GetDefaultGoogleOpOptions()
+	opOptions.SignGQ = signGQ
+  op := providers.NewGoogleOpWithOptions(opOptions)
 ```
 
 Next we create the OpenPubkey client and call `opkClient.Auth`:
--- a/README.md
+++ b/README.md
@@ -24,9 +24,9 @@ Let's walk through a simple message signing example. For conciseness we omit the
 We start by configuring the OP (OpenID Provider) our client and verifier will use. In this example we use Google as our OP.
 
 ```golang
-	opOptions := providers.GetDefaultGoogleOpOptions()
-	opOptions.SignGQ = signGQ
-  op := providers.NewGoogleOpWithOptions(opOptions)
+opOptions := providers.GetDefaultGoogleOpOptions()
+opOptions.SignGQ = signGQ
+op := providers.NewGoogleOpWithOptions(opOptions)
 ```
 
 Next we create the OpenPubkey client and call `opkClient.Auth`:
--- a/providers/google.go
+++ b/providers/google.go
@@ -43,7 +43,7 @@ var (
 
 const googleIssuer = "https://accounts.google.com"
 
-type Options struct {
+type GoogleOptions struct {
 	ClientID     string
 	ClientSecret string
 	Issuer       string // This should almost always be "https://accounts.google.com"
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -24,14 +24,9 @@ Let's walk through a simple message signing example. For conciseness we omit the
...
-  ClientID:     "992028499768-ce9juclb3vvckh23r83fjkmvf1lvjq18.apps.googleusercontent.com",
-  ClientSecret: "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", // The client secret is a public value
-  Scopes:       []string{"openid profile email"},
-  RedirURIPort: "3000",
-  CallbackPath: "/login-callback",
-  RedirectURI:  "http://localhost:3000/login-callback",
-}
+	opOptions := providers.GetDefaultGoogleOpOptions()
+	opOptions.SignGQ = signGQ
+  op := providers.NewGoogleOpWithOptions(opOptions)
```
EthanHeilman, line 29:
```suggestion
opOptions := providers.GetDefaultGoogleOpOptions()
opOptions.SignGQ = signGQ
op := providers.NewGoogleOpWithOptions(opOptions)
```

Too much whitespace

providers/google.go
```diff
@@ -222,3 +265,22 @@ func (g *GoogleOp) VerifyProvider(ctx context.Context, idt []byte, cic *clientin
...
+	var ln net.Listener
+	var lnErr error
+	for _, v := range redirectURIs {
+		redirectURI, err := url.Parse(v)
+		if err != nil {
+			return nil, nil, fmt.Errorf("malformed redirectURI specified, redirectURI was %s", v)
+		}
+
+		lnStr := fmt.Sprintf("localhost:%s", redirectURI.Port())
+		ln, lnErr = net.Listen("tcp", lnStr)
```
ymarcus93, line 280: We use `127.0.0.1` explicitly as the host instead of `localhost` in the listener string in `opk-ssh`, not sure if that makes any difference though given that `localhost` should map to that

providers/google.go
```diff
@@ -133,25 +181,20 @@ func (g *GoogleOp) requestTokens(ctx context.Context, cicHash string) ([]byte, e
...
-		err := g.server.ListenAndServe()
+		err := g.server.Serve(ln)
 		if err != nil && err != http.ErrServerClosed {
 			logrus.Error(err)
 		}
 	}()
 
+	loginURI := fmt.Sprintf("http://localhost:%s/login", redirectURI.Port())
+	logrus.Infof("Opening browser to on http://%s/", loginURI)
+	util.OpenUrl(loginURI)
```
ymarcus93, line 196: I might make a PR after this one is merged to add an option in `provider.Options` to disable automatically opening the browser (useful for integration tests). After that, the only difference in `opk-ssh`'s `GoogleOp` is refresh functionality I believe, so perhaps we can get closer to reusing the one here in `openpubkey` without having to duplicate the same core logic.

providers/google.go
```diff
@@ -222,3 +265,22 @@ func (g *GoogleOp) VerifyProvider(ctx context.Context, idt []byte, cic *clientin
...
+			return nil, nil, fmt.Errorf("malformed redirectURI specified, redirectURI was %s", v)
+		}
+
+		lnStr := fmt.Sprintf("localhost:%s", redirectURI.Port())
+		ln, lnErr = net.Listen("tcp", lnStr)
+		if lnErr == nil {
+			return redirectURI, ln, nil
+		}
+	}
+	return nil, nil, fmt.Errorf("failed to start a listener for the callback from Google, got %w", lnErr)
```
ymarcus93, line 285: nit: I don't think you need to include `Google` in the error string; this logic is useful for other OPs too and I don't think the logic here is specific to Google

providers/google.go
EthanHeilman: I'm using localhost here because localhost is used in the redirectURI configured in Google. I don't think it should make any difference as this repo has always used localhost in this way.

providers/google.go
EthanHeilman: Created an issue for this
https://github.com/openpubkey/openpubkey/issues/169

providers/google.go
EthanHeilman: Good point, fixed!

ymarcus93 (approved):

mrjoelkamp (approved):

Issue #168, opened 2024-04-04
SignGQ and GQSign Confusion


EthanHeilman:
For reasons that are entirely my fault, we use the terms SignGQ and GQSIgn interchangably in the codebase. This is confusing. We should switch to just using GQSign.

PR #170, opened 2024-04-05
Release 0.3 cleanup 

EthanHeilman:
This PR performs three cleanup needed for release 0.3.

1. Solves the inconsistent naming of boolean flag for turning on GQ Signing. In some places we say SignGQ in other places we say GQSign. We rename such flags to GQSIgn. Fixes: #168 
2. Renames renames the function which verifies an ID Token from `.VerifyProvider(...)` to `VerifyIDToken(...)`. I'm normally against renames like this but given that the name of the struct is `ProviderVerifier` having the function be `ProviderVerifier.VerifyProvider(...)` costs me 10 minutes doing the wrong thing too many times. Given that the function for verifying a PK Token is `VerifyPKToken(...)` it seems nature that the function for verifying an ID Token would be `VerifyIDToken(...)`.
3. Deletes unused compatibility function. Release 0.3 will have many breaking changes, it is time to retire `client.VerifyPKToken(...)` and just use `verifier.VerifyPKToken(...)`

## Tests

- [X] github test https://github.com/openpubkey/gha-test/actions/runs/8573505646/job/23498328408
- [X] gitlab-ci test https://gitlab.com/openpubkey/gl-test/-/pipelines/1241996045
- [X] MFA Cosigner test (on OSX/brave)
- [X] Simple example (os OSX/brave, OSX/firefox, OSX/Safari, OSX/chrome)

mfa cosigner test


mrjoelkamp (approved):
LGTM

Issue #166, opened 2024-04-04
Compact Representation Format for PK Tokens


EthanHeilman:
This issue proposes a Compact representation for PK Tokens. The goals of such a format are:
* Must avoid multiple base64 encodes to result minimal size, that is a compact representation,
* Must handle including JKT of OP public key, if one exists
* Must handle refreshed ID Tokens
* Must not include unsigned values

## Proposed Format
Our compact representation format for PK Token:

```golang
Base64(payload)+":"+Base64(OP protected)+":"+Base64(OP sig)+":"+Base64(CIC Protected)+":"+Base6464(CIC Sig)
```
If we wish to include a refreshed ID Token, we add it to the back using the JWT compact representation: 
```golang
Base64(refreshed-protected)+"."+Base64(refreshed-payload)+"."+Base64(refreshed-sig)
```
and get (newlines added for clarity):

```golang
Base64(payload)+":"
+Base64(OP protected)+":"+Base64(OP sig)+":"
+Base64(CIC protected)+":"+Base64(CIC sig)+"."
+Base64(refreshed-protected)+"."+Base64(refreshed-payload)+"."+Base64(refreshed-sig)
```
By protected, we denote the protected header that is paired with a particular signature.

The order of protected header, signature pairs doesn't matter since the `typ` claim in each protected allows anyone to determine the type of protected header: is it a OP, CIC or COS protected header and signature. 

The above example only shows the OP and CIC protected headers and signatures, but this compact format supports an unlimited number of protected header and signature pairs.

```golang
Base64(payload)+":"
+Base64(protected_1)+":"+Base64(sig_1)+":"
+Base64(protected_2)+":"+Base64(sig_2)+":"
+Base64(protected_3)+":"+Base64(sig_3)+":"
// ...
+Base64(protected_N)+":"+Base64(sig_N)+"."
+Base64(refreshed-protected)+"."+Base64(refreshed-payload)+"."+Base64(refreshed-sig)
```

## Parsing

This compact representation is intended to be easy to parse.

```
compact_PKT, compact_RIDT.protected, compact_RIDT.payload, compact_RIDT.sig=Split(".", compact_PKT)

parts=Split(":", compact_PKT)
compact_PKT.payload=parts[0]
compact_PKT.OP.protected=parts[1]
compact_PKT.OP.sig=parts[2]
compact_PKT.CIC.protected=parts[3]
compact_PKT.CIC.sig=parts[4]
```

## The JKT field

[...]

PR #177, opened 2024-04-11
feat: Adds JKT (thumprint) to Protected Header of GQ Signed ID Tokens

### LINKED ISSUE #172: JKT in GQ ID Token
Now that we have compact representations of PK Tokens, we should add a way to represent JKT in compact PK Tokens. 

The current plan for doing this was discussed in https://github.com/openpubkey/openpubkey/issues/166#issuecomment-2040348944

> When replacing the RSA signature with a GQ Signature, we create a new protected header that contains the original RSA protected header in kid. We could easily just add the jtk to the new GQ protected header.

EthanHeilman:
OpenID Providers (OP) are not required to set kids (Key IDs) or use unique kids for the keys in their JWKS. It is extremely rare for an OP to not use an kid or to recycle a kid but it could happen. To ensure we can always find the correct public key to verify a ID Token, we record the JKT (JSON Key Thumbprint) of the OP's public key.

Currently when we serialize a PK Token as a JSON, we add the JTK to a public header. Unfortunately this approach does not work for compact representation of  PK Token because we want the compact representation only include signed data and public headers are not signed.

I made an realization that we only really care about JKTs for archival verification of messages signed my PK Tokens. When PK Tokens are used for signatures we always replay the RSA signature on the ID Token with a GQ signature (or in the future a ZKP signature). Thus, we only need to solve this problem for GQ signed PK Tokens. So why not just put the JKT claim in the GQ protected header of the ID Token? This solves all our problems with JKT!

The GQ signature covers the protected header so the JKT is now signed. Even better the GQ protected header is already included in the compact representation so we don't have to change the compact representation.

This PR puts the JKT in the protected header of the GQ Signature using under the key "jkt". This PR does not remove the code which adds the JKT to the PK Token when not using GQ Signatures. I'm of the opinion that it should remove it, but I want to discuss this with @jonnystoten first. I've created the issue to defer #178  and track this decision.

Fixes #172 

This solution was originally proposed here: https://github.com/openpubkey/openpubkey/issues/166#issuecomment-2040348944

## Tests

- [X] Github Actions manual test https://github.com/openpubkey/gha-test/actions/runs/8655405581/job/23734297610
- [X] Gitlab manual test https://gitlab.com/openpubkey/gl-test/-/pipelines/1250028405
- [X] MFA Cosigner manual test f9a1010d28907a1959e0562bd369a2e3522b64ea

### DIFF
--- a/providers/gq_test.go
+++ b/providers/gq_test.go
@@ -73,7 +73,7 @@ func TestGQ(t *testing.T) {
 
 			op, backend, idtTemplate, err := NewMockProvider(providerOpts)
 			require.NoError(t, err)
-			
+
 			expPublicKey := maps.Values(backend.GetProviderPublicKeySet())[0].PublicKey
 
 			idToken, err := idtTemplate.IssueToken()
--- a/providers/gq_test.go
+++ b/providers/gq_test.go
@@ -72,8 +72,8 @@ func TestGQ(t *testing.T) {
 			providerOpts.CommitType = tc.tokenCommitType
 
 			op, backend, idtTemplate, err := NewMockProvider(providerOpts)
-
 			require.NoError(t, err)
+			
 			expPublicKey := maps.Values(backend.GetProviderPublicKeySet())[0].PublicKey
 
 			idToken, err := idtTemplate.IssueToken()
--- a/providers/gq_test.go
+++ b/providers/gq_test.go
@@ -1,0 +1,119 @@
+// Copyright 2024 OpenPubkey
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//	http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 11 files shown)


providers/gq_test.go
```diff
@@ -0,0 +1,119 @@
...
+		t.Run(tc.name, func(t *testing.T) {
+
+			providerOpts := DefaultMockProviderOpts()
+			providerOpts.NumKeys = 1 // Only use 1 public key so we can find the public key used
+			providerOpts.CommitType = tc.tokenCommitType
+
+			op, backend, idtTemplate, err := NewMockProvider(providerOpts)
+
+			require.NoError(t, err)
+			expPublicKey := maps.Values(backend.GetProviderPublicKeySet())[0].PublicKey
```
mrjoelkamp, line 77:
```suggestion
			op, backend, idtTemplate, err := NewMockProvider(providerOpts)
			require.NoError(t, err)
			
			expPublicKey := maps.Values(backend.GetProviderPublicKeySet())[0].PublicKey
```

mrjoelkamp (approved):
the extra claims opts come in nicely here! minor nit on spacing, otherwise LGTM

PR #181, opened 2024-04-13
test: Migrated the rest of the test to testify/require

### LINKED ISSUE #175: Move all tests over to testify/require
Currently some tests aren't using the require pattern as shown below:

https://github.com/openpubkey/openpubkey/blob/57897d7475b79ec089285b3267598dab3fdd0153/examples/ssh/opkssh/cli_test.go#L55-L58

We should move all of the remaining tests over to require:

https://github.com/openpubkey/openpubkey/blob/03c27ef14492728c0125c3691b35b1a79b3d1e09/client/opkclient_test.go#L78-L81

roceb:
Fixes #175 
Moved remaining test to use testify/require

### DIFF
--- a/gq/benchmark_test.go
+++ b/gq/benchmark_test.go
@@ -19,8 +19,9 @@ package gq
 import (
 	"crypto/rand"
 	"crypto/rsa"
-	"fmt"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 var result error
@@ -34,9 +35,7 @@ type testTuple struct {
 func BenchmarkSigning(b *testing.B) {
 	// Generate test matrix
 	matrix, err := generateTestMatrix(b.N)
-	if err != nil {
-		b.Fatal(err)
-	}
+	require.NoError(b, err)
 
 	// Reset the benchmark timer to exclude setup time
 	b.ResetTimer()
@@ -44,10 +43,9 @@ func BenchmarkSigning(b *testing.B) {
 	var signerVerifier SignerVerifier
 	for i := 0; i < b.N; i++ {
 		signerVerifier, err = NewSignerVerifier(matrix[i].rsaPublicKey, 256)
-		if err != nil {
-			b.Fatal(err)
-		}
+		require.NoError(b, err)
 		_, err = signerVerifier.SignJWT(matrix[i].token)
+		require.NoError(b, err)
 	}
 
 	// Avoid compiler optimisations eliminating the function under test and artificially lowering the run time of the benchmark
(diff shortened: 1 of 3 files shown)


gq/benchmark_test.go
```diff
@@ -34,20 +35,17 @@ type testTuple struct {
...
 
 	var signerVerifier SignerVerifier
 	for i := 0; i < b.N; i++ {
 		signerVerifier, err = NewSignerVerifier(matrix[i].rsaPublicKey, 256)
-		if err != nil {
-			b.Fatal(err)
-		}
+		require.NoError(b, err)
 		_, err = signerVerifier.SignJWT(matrix[i].token)
+		require.NoError(b, err)
```
roceb, line 48: The error for this test was originally ignored, not sure if this was on purpose but I handled it for now. Please advise whether to keep it or remove it

providers/providerverifier_test.go
```diff
@@ -114,12 +114,14 @@ func TestProviderVerifier(t *testing.T) {
 				NoAlg:   false,
 			}
 
-			if tc.tokenCommitType.Claim == "nonce" {
+			switch tc.tokenCommitType.Claim {
```
roceb, line 117: Switched if statement to a case statement. This wasn't part of the issue but it cleans up the code a bit

gq/benchmark_test.go
EthanHeilman: I think it is worth checking just in case and to reduce the number of warnings.

providers/providerverifier_test.go
EthanHeilman: I like it and makes the code more readable

gq/gq_iso_test.go
```diff
@@ -94,9 +88,7 @@ func TestVerifierISO(t *testing.T) {
 	encodedSigISO := util.Base64EncodeForJWT(sigISO)
 	ok := svISO.Verify(encodedSigISO, idISO, mISO)
 
-	if !ok {
-		t.Fatal("signature verification failed")
-	}
+	require.True(t, ok, "signature verification failed")
```
EthanHeilman, line 91: Since you are touching this code, might as well fix the lowercase message. I looked it up because I wasn't sure if require messages should should start with uppercase or lowercase and they are uppercase.

```suggestion
	require.True(t, ok, "Signature verification failed")
```

EthanHeilman (approved):
Thanks for cleaning this code up!

EthanHeilman:
If you are looking for something to else to work on.

The verifier test is  has a lot of duplication and could use a clean up and better structure
https://github.com/openpubkey/openpubkey/blob/main/verifier/verifier_test.go

Also I've been using the mock provider in a lot of tests and adding features and options to it and it has gotten pretty awkward. It needs some care and re-engineering:
https://github.com/openpubkey/openpubkey/blob/main/providers/mockprovider.go

roceb:
I can take a look at that

PR #188, opened 2024-04-17
SPDX-License-Identifier: Apache-2.0. fix #185

### LINKED ISSUE #185: We should use algorithm from public key rather than alg from token in verification
In the code below, we use the alg set in the ID Token to verify the ID Token. This does not present a security risk because we also check it is the expected algorithm and for GQ256 we need to do this. However we should check that public key in `pubKeyRecord.PublicKey` is RS256. 

```go
	algStr := idt.GetSignature().GetProtectedClaims().Alg
	if algStr == "" {
		return fmt.Errorf("provider algorithm type missing")
	}
	alg := jwa.SignatureAlgorithm(algStr)
	if alg != gq.GQ256 && v.options.GQOnly {
		return fmt.Errorf("non-GQ signatures are not supported")
	}

	switch alg {
	case gq.GQ256:
		if err := v.verifyGQSig(ctx, idt); err != nil {
			return fmt.Errorf("error verifying OP GQ signature on PK Token: %w", err)
		}
	case jwa.RS256:
		pubKeyRecord, err := v.providerPublicKey(ctx, idToken)
		if err != nil {
			return fmt.Errorf("failed to get OP public key: %w", err)
		}
		alg := jwa.SignatureAlgorithm(algStr)
		if _, err := jws.Verify(idToken, jws.WithKey(alg, pubKeyRecord.PublicKey)); err != nil {
			return err
		}
	}
```

thebigbone:


providers/providerverifier.go
```diff
@@ -132,6 +132,11 @@ func (v *DefaultProviderVerifier) VerifyIDToken(ctx context.Context, idToken []b
 			return fmt.Errorf("failed to get OP public key: %w", err)
 		}
 
+		_, ok := pubKeyRecord.PublicKey.(*rsa.PublicKey)
```
EthanHeilman, line 135: Nit: added a clarifying comment here 
```suggestion
        // Ensure that the algorithm of public key from OpenID Provider matches 
        // the algorithm specified in the ID Token
		_, ok := pubKeyRecord.PublicKey.(*rsa.PublicKey)
```

EthanHeilman:
LGTM to merge, just accept the suggestion or rewrite the comment

thebigbone:
I have added the comment. Thanks for the suggestion.

EthanHeilman (approved):
Thanks for this bug fix

providers/providerverifier.go
```diff
@@ -132,6 +132,12 @@ func (v *DefaultProviderVerifier) VerifyIDToken(ctx context.Context, idToken []b
 			return fmt.Errorf("failed to get OP public key: %w", err)
 		}
 
+        // Ensure that the algorithm of public key from OpenID Provider matches the algorithm specified in the ID Token
```
EthanHeilman, line 135: Linter is angry because tabs instead of spaces

```suggestion
		// Ensure that the algorithm of public key from OpenID Provider matches the algorithm specified in the ID Token
```

PR #194, opened 2024-04-18
Renames oidc package to jwsig

RishavKumarSinha:


EthanHeilman:
Use the `git mv` command to preserve the git history

jwsig/jws.go
```diff
@@ -0,0 +1,114 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+package jwsig
```
EthanHeilman, line 17: Thanks for putting this together

Use the `git mv` command to move the files. This will preserve the git history

RishavKumarSinha: Got it, but what will be the source file name and destination file in the git mv command? Is it just `git mv oidc jwsig`?

EthanHeilman: Yeah, that should work. If it doesn't checkout https://git-scm.com/docs/git-mv

RishavKumarSinha: @EthanHeilman, I tried following the docs but I am still ending up with 
```
$ git mv oidc jwsig
fatal: bad source, source=oidc, destination=jwsig/oidc
```
Can you help me figure out why this persists?

EthanHeilman: what directory are you?

RishavKumarSinha: jwsig

PR #196, opened 2024-04-23
Add more options to `GoogleOp` provider + Misc bug fixes

### LINKED ISSUE #169: Add option in provider.Options to disable opening the browser
Add an option in provider.Options to disable automatically opening the browser (useful for integration tests). After that, the only difference in opk-ssh's GoogleOp is refresh functionality I believe, so perhaps we can get closer to reusing the one here in openpubkey without having to duplicate the same core logic.

From:
https://github.com/openpubkey/openpubkey/pull/167#pullrequestreview-1983298265

ymarcus93:
Adds more options to `GoogleOp` provider, and fixes some miscellaneous bugs found when upgrading `opk-ssh` to latest commit upstream in `openpubkey`.

+ New options in `GoogleOp`:
  + `OpenBrowser` option. Defaults to `true`. If `false`, then the browser window does not open when performing OIDC auth with the `GoogleOp` provider.
  + `HttpClient` option. Defaults to `nil`. If non-nil, then provided `HttpClient` is used when making queries to the OP (OIDC code exchange, refresh, verification of ID token, fetch of JWKS endpoint). Otherwise, `http.DefaultClient` is used as before.
  + `IssuedAtOffset` option. Defaults to `1 * time.Minute`. Allows for customizing the offset/clock skew option passed to `zitadel` library when verifying id tokens' `iat` and `exp` claims.
+ Update `RefreshableOpenIdProvider` interface to embed `OpenIdProvider` as a `RefreshableOpenIdProvider` in itself should also be an `OpenIdProvider` (similar to type def of `BrowserOpenIdProvider`)
+ Pass `GoogleOp`'s `publicKeyFinder` to `NewProviderVerifier`
+ Fix issue in `RefreshTokens` where after first refresh the `RefreshToken` would be lost as Google does not return a refresh token
+ Remove double `http.Server.Serve()` in `requestTokens`
+ Create `NewServeMux` instead of using global mux in `http` package, so that you can run two instances of `GoogleOp` without conflicting registered handlers.
+ ~~Refactor `shutdownServer` logic~~
+ Fix bug in `NewFromCompact` not setting `FreshIDToken` when parsing compact representation of PKT

Fixes #169

TODO:
- [x] Run manual tests, especially cosigner flow, as I changed shutdown logic

## Tests
- [x] Manual MFA Cosigner test
- [X] Gitlab-ci test https://gitlab.com/openpubkey/gl-test/-/jobs/6699953337
- [X] Github Actions test https://github.com/openpubkey/gha-test/actions/runs/8807078974/job/24173315870

### DIFF
--- a/discover/discover.go
+++ b/discover/discover.go
@@ -76,21 +76,28 @@ func NewPublicKeyRecord(key jwk.Key, issuer string) (*PublicKeyRecord, error) {
 
 func DefaultPubkeyFinder() *PublicKeyFinder {
 	return &PublicKeyFinder{
-		JwksFunc: GetJwksByIssuer,
+		JwksFunc:   GetJwksByIssuer,
+		HttpClient: nil,
 	}
 }
 
-type JwksFetchFunc func(ctx context.Context, issuer string) ([]byte, error)
+type JwksFetchFunc func(ctx context.Context, issuer string, httpClient *http.Client) ([]byte, error)
 
 type PublicKeyFinder struct {
-	JwksFunc JwksFetchFunc
+	JwksFunc   JwksFetchFunc
+	HttpClient *http.Client
 }
 
-// GetJwksByIssuer fetches the JWKS from the issuer's JWKS endpoint found at the issuer's well-known
-// configuration. It doesn't attempt to parse the response but instead returns the JSON bytes of
-// the JWKS.
-func GetJwksByIssuer(ctx context.Context, issuer string) ([]byte, error) {
-	discConf, err := oidcclient.Discover(ctx, issuer, http.DefaultClient)
+// GetJwksByIssuer fetches the JWKS from the issuer's JWKS endpoint found at the
+// issuer's well-known configuration. It doesn't attempt to parse the response
+// but instead returns the JSON bytes of the JWKS. If httpClient is nil, then
+// http.DefaultClient is used when fetching.
+func GetJwksByIssuer(ctx context.Context, issuer string, httpClient *http.Client) ([]byte, error) {
+	if httpClient == nil {
+		httpClient = http.DefaultClient
+	}
+
+	discConf, err := oidcclient.Discover(ctx, issuer, httpClient)
 	if err != nil {
 		return nil, fmt.Errorf("failed to call OIDC discovery endpoint: %w", err)
 	}
@@ -100,13 +107,13 @@ func GetJwksByIssuer(ctx context.Context, issuer string) ([]byte, error) {
 		return nil, err
 	}
 
-	response, err := http.DefaultClient.Do(request)
+	response, err := httpClient.Do(request)
 	if err != nil {
 		return nil, err
 	}
 	defer response.Body.Close()
 
-	resp, err := http.DefaultClient.Get(discConf.JwksURI)
+	resp, err := httpClient.Get(discConf.JwksURI)
 	if err != nil {
 		return nil, fmt.Errorf("failed to fetch to JWKS: %w", err)
 	}
@@ -117,7 +124,7 @@ func GetJwksByIssuer(ctx context.Context, issuer string) ([]byte, error) {
 }
 
 func (f *PublicKeyFinder) fetchAndParseJwks(ctx context.Context, issuer string) (jwk.Set, error) {
-	jwksJson, err := f.JwksFunc(ctx, issuer)
+	jwksJson, err := f.JwksFunc(ctx, issuer, f.HttpClient)
 	if err != nil {
 		return nil, fmt.Errorf(`failed to fetch JWKS: %w`, err)
 	}
--- a/discover/discover.go
+++ b/discover/discover.go
@@ -76,16 +76,16 @@ func NewPublicKeyRecord(key jwk.Key, issuer string) (*PublicKeyRecord, error) {
 
 func DefaultPubkeyFinder() *PublicKeyFinder {
 	return &PublicKeyFinder{
-		JwksFunc:   GetJwksByIssuer,
-		HttpClient: nil,
+		JwksFunc: func(ctx context.Context, issuer string) ([]byte, error) {
+			return GetJwksByIssuer(ctx, issuer, nil)
+		},
 	}
 }
 
-type JwksFetchFunc func(ctx context.Context, issuer string, httpClient *http.Client) ([]byte, error)
+type JwksFetchFunc func(ctx context.Context, issuer string) ([]byte, error)
 
 type PublicKeyFinder struct {
-	JwksFunc   JwksFetchFunc
-	HttpClient *http.Client
+	JwksFunc JwksFetchFunc
 }
 
 // GetJwksByIssuer fetches the JWKS from the issuer's JWKS endpoint found at the
@@ -124,7 +124,7 @@ func GetJwksByIssuer(ctx context.Context, issuer string, httpClient *http.Client
 }
 
 func (f *PublicKeyFinder) fetchAndParseJwks(ctx context.Context, issuer string) (jwk.Set, error) {
-	jwksJson, err := f.JwksFunc(ctx, issuer, f.HttpClient)
+	jwksJson, err := f.JwksFunc(ctx, issuer)
 	if err != nil {
 		return nil, fmt.Errorf(`failed to fetch JWKS: %w`, err)
 	}
--- a/pktoken/pktoken.go
+++ b/pktoken/pktoken.go
@@ -106,7 +106,7 @@ func New(idToken []byte, cicToken []byte) (*PKToken, error) {
 
 // NewFromCompact creates a PK Token from a compact representation
 func NewFromCompact(pktCom []byte) (*PKToken, error) {
-	tokens, _, err := SplitCompactPKToken(pktCom)
+	tokens, freshIDToken, err := SplitCompactPKToken(pktCom)
 	if err != nil {
 		return nil, err
 	}
@@ -123,6 +123,7 @@ func NewFromCompact(pktCom []byte) (*PKToken, error) {
 			return nil, err
 		}
 	}
+	pkt.FreshIDToken = freshIDToken
 	return pkt, nil
 }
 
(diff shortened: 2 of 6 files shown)


discover/discover.go
```diff
@@ -76,21 +76,28 @@ func NewPublicKeyRecord(key jwk.Key, issuer string) (*PublicKeyRecord, error) {
...
 	}
 }
 
-type JwksFetchFunc func(ctx context.Context, issuer string) ([]byte, error)
+type JwksFetchFunc func(ctx context.Context, issuer string, httpClient *http.Client) ([]byte, error)
 
 type PublicKeyFinder struct {
-	JwksFunc JwksFetchFunc
+	JwksFunc   JwksFetchFunc
+	HttpClient *http.Client
```
EthanHeilman, line 88: Do we need both of these?

If JwksFetchFunc takes httpClient as a parameter, then do we need HttpClient on the struct?
Alternatively if HttpClient is on the struct couldn't it just get HttpClient via a closure?

pktoken/pktoken.go
```diff
@@ -106,7 +106,7 @@ func New(idToken []byte, cicToken []byte) (*PKToken, error) {
 
 // NewFromCompact creates a PK Token from a compact representation
 func NewFromCompact(pktCom []byte) (*PKToken, error) {
-	tokens, _, err := SplitCompactPKToken(pktCom)
+	tokens, freshIDToken, err := SplitCompactPKToken(pktCom)
```
EthanHeilman, line 109: Thanks for fixing this

providers/google.go
```diff
@@ -39,19 +39,18 @@ import (
...
-	Scopes       []string
-	RedirectURIs []string
-	GQSign       bool
+	ClientID       string
+	ClientSecret   string
+	Issuer         string // This should almost always be "https://accounts.google.com"
+	Scopes         []string
+	RedirectURIs   []string
+	GQSign         bool
+	OpenBrowser    bool
```
EthanHeilman, line 51: Can you add a comment here explaining some of these new struct variables? For instance no one should set OpenBrowser = false outside of integration tests

providers/google.go
```diff
@@ -60,6 +59,9 @@ type GoogleOp struct {
 	Scopes                    []string
 	RedirectURIs              []string
 	GQSign                    bool
+	OpenBrowser               bool
```
EthanHeilman, line 62: why is OpenBrowser exposed but httpClient, not exposed. They should probably both be not exposed or is there some value is setting them directly?

providers/google.go
```diff
@@ -96,21 +100,28 @@ func NewGoogleOp() OpenIdProvider {
...
 		publicKeyFinder:           *discover.DefaultPubkeyFinder(),
 	}
+
+	googleOp.publicKeyFinder.HttpClient = opts.HttpClient
+	return googleOp
 }
 
 var _ OpenIdProvider = (*GoogleOp)(nil)
 var _ BrowserOpenIdProvider = (*GoogleOp)(nil)
+var _ RefreshableOpenIdProvider = (*GoogleOp)(nil)
```
EthanHeilman, line 124: Nice!

providers/google.go
```diff
@@ -124,13 +135,8 @@ func (g *GoogleOp) requestTokens(ctx context.Context, cicHash string) (*simpleoi
 	logrus.Infof("listening on http://%s/", ln.Addr().String())
 	logrus.Info("press ctrl+c to stop")
 
-	g.server = &http.Server{}
```
EthanHeilman, line 127: Give that we say we are listening and just found the port. Is there a good reason not to start the server here? Doesn't starting the server here eliminate any chance of race conditions where the server isn't running yet but we have triggered the callback to the server?

providers/google.go
```diff
@@ -185,51 +189,46 @@ func (g *GoogleOp) requestTokens(ctx context.Context, cicHash string) (*simpleoi
...
-	http.Handle(callbackPath, rp.CodeExchangeHandler(marshalToken, relyingParty))
+	mux.Handle(callbackPath, rp.CodeExchangeHandler(marshalToken, relyingParty))
 
 	go func() {
 		err := g.server.Serve(ln)
 		if err != nil && err != http.ErrServerClosed {
 			logrus.Error(err)
 		}
 	}()
+	defer func() {
```
EthanHeilman, line 216: I don't think this approach works with the cosigner.
1. requestTokens exits, defer called, server shuts down,
2. cosigner flow starts and calls HookHTTPSession
3. HookHTTPSession attempts to do the redirect using the server but the server is shutdown.  Error!

I think this breaks the cosigner flow

EthanHeilman (changes requested):
Unless I am misunderstanding something about your changes, I believe this breaks the cosigner flow. 

I think if you didn't change the shutdown logic, it wouldn't break.

If I am wrong and this doesn't break the cosigner flow, then the shutdown logic design is great and simplifies things a lot

providers/google.go
ymarcus93: Interesting. For some reason, I thought everything was fine because previous code looks like this:

https://github.com/openpubkey/openpubkey/blob/eba749e7d12b83798f2ab29ed684ba3bafed5e6f/providers/google.go#L190-L200

the `hook` is called and then once `hook` func returns `serverShutdown()` is called immediately (this is in `callbackPath` handler).

With the changes in this PR, I flipped the order so that the hook happens first and then we write to `chTokens` channel

https://github.com/ymarcus93/openpubkey/blob/878cab32fd7555dde1a0b55dd8fa16a65742bbae/providers/google.go#L192-L204

Once `chTokens` is written to `requestToken` exits and `defer` is called and server shuts down. But by that point shouldn't the `hookHttpSession` method already have been executed?

providers/google.go
ymarcus93: I kept the `Serve` in the original location because it is called after all the handlers are registered on the `mux` handler: `/login` and `callbackPath`. Then immediately after `Serve`, `OpenBrowser` logic occurs. If we move it to here before the handlers are registered, then there is a race as the handlers haven't been registered yet.

providers/google.go
ymarcus93: Good point. Yeah I should be consistent here and expose it. Not sure why I didn't do that.

providers/google.go
ymarcus93: I guess my question is when does the cosigner flow start? When `httpSessionHook` is called or after `requestToken` exits?

discover/discover.go
ymarcus93: Thanks! Fixed here: be253b5

providers/google.go
EthanHeilman: Cosigner flow:
1. Cosigner sets httpSessionHook hook,
2. oidcAuth called
  3. Oidc Auth calls OP.requestTokens
  4. OP.requestTokens finishes
  5. oidcAuth finishes
6. Cosigner flow starts
7. Cosigner flow uses hook to redirect user's open webpage to cosigner init auth
8. After init auth completes, OP http server is shutdown
9. Cosigner flow finishes

EthanHeilman: The line
```go
 if g.httpSessionHook != nil { 
 	g.httpSessionHook(w, r) 
 	defer shutdownServer() // If no http session hook is set, we do server shutdown in RequestTokens 
 ```
 
 Is happening inside marshalToken after `chTokens <- retTokens` and `chTokens <- retTokens` causes requestTokens to return.

Comment (EthanHeilman) [reply providers/google.go:217]: > I guess my question is when does the cosigner flow start? When httpSessionHook is called or after requestToken exits?

The cosigner code starts when requestToken exits, but the flow is kicked off by using the channel in httpSessionHook to redirect the user's web browser

1. Client.Auth sets the hook

https://github.com/openpubkey/openpubkey/blob/eba749e7d12b83798f2ab29ed684ba3bafed5e6f/client/client.go#L188C1-L200C4

2. Client.Auth then calls requestTokens

3. requestTokens does the OIDC flow and returns, but marshalTokens is still blocking on the hook

4. After requestTokens returns, the Cosigner flow starts,

5. The Cosigner flow uses the hook to redirect the http server and unblocks marshalTokens

I dislike how complex this is, but I haven't found a better way of doing it.

File: providers/google.go (hunk 3 above)
Comment (ymarcus93) [reply providers/google.go]: Comments added here: 53cbe3e

File: providers/google.go (hunk 4 above)
Comment (ymarcus93) [reply providers/google.go:89]: Fixed here: 58f9751

**Comment by EthanHeilman** [comment]
I don't see any reason not to merge, but I'm going to run some tests and stare really hard at the GoogleOP changes. Likely we can get this merged before sunset

**Comment by EthanHeilman** [comment]
I tested the MFA-cosigner and the google example with refresh flow on windows 11/brave successfully against commit 58f9751ec46e87385f1f1cac390fabe276033f80

File: providers/google.go [hunk 8]
```diff
@@ -124,13 +163,8 @@ func (g *GoogleOp) requestTokens(ctx context.Context, cicHash string) (*simpleoi
...
 
-	g.server = &http.Server{}
-	go func() {
-		err := g.server.Serve(ln)
-		if err != nil && err != http.ErrServerClosed {
-			logrus.Error(err)
-		}
-	}()
+	mux := http.NewServeMux()
+	g.server = &http.Server{Handler: mux}
```
EthanHeilman, line 167: Since `g.server.Serve(ln)` now happens after `defer shutdownServer()` what happens if an error causes the function to return before running `g.server.Serve(ln)`. Will this cause a nil panic since the server isn't running or is this safe because you did initialize `g.server`?

EthanHeilman (approved):
LGTM, just comment that you checked my one question isn't an issue

providers/google.go
ymarcus93: `g.server` is set before the `shutdownServer` closure is defined, so I don't think we'll get a nil panic even if the server isn't running (also according to `Shutdown()` it makes future calls on `Serve` return `ErrServerClosed`).

PR #204, opened 2024-05-08
Deletes unused PKToken struct function Sign

EthanHeilman:
The function Sign is not used anywhere. This PR deletes it.


jonnystoten (approved):
LGTM :broom:

PR #205, opened 2024-05-10
test: add test for op.go

thebigbone:
Adds simple env variable test for providers package

### DIFF
--- a/providers/op_test.go
+++ b/providers/op_test.go
@@ -1,0 +1,25 @@
+package providers
+
+import (
+	"os"
+	"testing"
+)
+
+func TestGetEnvVar(t *testing.T) {
+	originalEnv := os.Getenv("TEST_ENV_VAR")
+	os.Setenv("TEST_ENV_VAR", "test value")
+	defer os.Setenv("TEST_ENV_VAR", originalEnv)
+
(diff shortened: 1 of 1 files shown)


providers/op_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+func TestGetEnvVar(t *testing.T) {
+	originalEnv := os.Getenv("TEST_ENV_VAR")
+	os.Setenv("TEST_ENV_VAR", "test value")
+	defer os.Setenv("TEST_ENV_VAR", originalEnv)
+
+	value, err := getEnvVar("TEST_ENV_VAR")
+	if err != nil {
+		t.Errorf("getEnvVar returned an error: %v", err)
+	}
```
EthanHeilman, line 16: We use `"github.com/stretchr/testify/require"` for tests, such as require.NoError(t, err) see https://github.com/openpubkey/openpubkey/blob/main/gq/gq_test.go

EthanHeilman (changes requested):

providers/op_test.go
thebigbone: Ops, I should have seen that. Will change it.

EthanHeilman (approved):

Issue #173, opened 2024-04-11
Docs on PK Token


EthanHeilman:
Doc should cover

1. Motivation behind PK Tokens
2. Terminology
3. How the typ protected header claim is used to distinguish different types of signatures on a PK Token
4. Representations: JSON, Compact
5. GQ ID Tokens
6. GQ Bound PK Tokens

PR #209, opened 2024-07-14
docs: Markdown doc to explain PK Tokens

EthanHeilman:
The intention here is to document all the different patterns we use in PK Tokens. This is less an introduction to OpenPubkey and more a reference or spec for the different PK Tokens.

This is a documentation only PR.

## TODOs

- [X]  Google Example PK Token
- [X]  Github Example PK Token
- [X]  Gitlab Example PK Token
- [X] Explain GQ replay protection (jtk)
- [X] Explain GQ bound PK Tokens
- [X] Explain ZK bound PK Tokens
- [X] Proof read 

Fixes https://github.com/openpubkey/openpubkey/issues/173


docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
+# A Guide to PK Tokens
+
+OpenPubkey works by binding an public key to an identity described in an ID Token. To provide verifiers with all the needed information to verify this binding between identity and public key we include additional information by extending the ID Token with additional signatures. This is possible because ID Tokens are JSON Web Signatures (JWS) and JWS can support more than one signature. We call an ID Token extended in this fashion a PK Token (Public Key Token).
```
asamborski, line 3: Spell check: "a public key" instead of "an"

asamborski, line 3: Suggested wording change: "This is possible because ID Tokens are JSON Web Signatures (JWS) and can support multiple signatures."

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+In this document we provide needed background on JSON Web Signatures (JWS), ID Tokens and PK Tokens. Then we describe how our different types of PK Tokens work using examples. In the next section we provide our compact serialization format. Finally we provide a short guide to using the patterns we developed for generic multi-signature JWS.
+
+- [A Guide to PK Tokens](#a-guide-to-pk-tokens)
+  - [JSON Web Signatures (JWS) and JSON Web Tokens (JWT)](#json-web-signatures-jws-and-json-web-tokens-jwt)
+  - [PK Tokens](#pk-tokens)
+    - [Protected Header Claims](#protected-header-claims)
+      - [Required Claims](#required-claims)
+      - [Custom Claims](#custom-claims)
+    - [Signature Type (typ)](#signature-type-typ)
+    - [Commitment mechanism](#commitment-mechanism)
```
asamborski, line 14: For consistency, may want to capitalize M in mechanism

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+      - [GQ Signatures Are Required For Aud-Commitment PK Tokens](#gq-signatures-are-required-for-aud-commitment-pk-tokens)
+    - [GQ-Commitment PK Tokens (Gitlab-CI Example)](#gq-commitment-pk-tokens-gitlab-ci-example)
+      - [GQ-Commitment to the CIC](#gq-commitment-to-the-cic)
+      - [Security](#security)
+    - [ZK PK Tokens (Under development)](#zk-pk-tokens-under-development)
+  - [PK Token Compact Serialization](#pk-token-compact-serialization)
+    - [Refreshed Payload and Signature](#refreshed-payload-and-signature)
+- [Notes](#notes)
+  - [GQ Signatures](#gq-signatures-1)
+  - [Our JWS conventions](#our-jws-conventions)
```
asamborski, line 38: Similar feedback - may want to capitalize these headers to be consistent with the top portion

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+    "protected": {"alg": "RS256", "kid": "9123"},
+    "signature": "signature-3"
+  },
+]}
+```
+
+
+Note that each signature in a JWS signs the payload and the protected header associated with that signature. All signatures sign the same payload; no signature signs another signature's protected header. In the example above RSA signature-2 is computed as  `RSA-SIGN(SK, ("message payload", {"alg": "RS256", "kid": "1234"}))`.
+
+[JWT (JSON Web Token)](https://datatracker.ietf.org/doc/html/rfc7519) is a type of JWS used by one party, the issuer, to make claims another parties, the subject. The issuer includes their identity in the JWT using the `iss` claim. JWTs are defined as having only one signature, the signature of the issuer.
```
asamborski, line 72: Did you mean to say, "is a type of JWS used by one party, the issuer, to make claims **against** another party, the subject" ?

asamborski: Or maybe you meant "about" ?

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+      "typ": "JWT"
+    },
+    "signature": "GqjU... (Issuer's signature)"
+  }
+]}
+```
+
+## PK Tokens
+
+Now let's look at what happens when we extend this ID Token with additional signatures to create a PK Token. A PK Token will have at least have at least two signatures but can have more than two signatures.
```
asamborski, line 111: Nix repeat wording: "A PK Token will have at least two signatures but can have more than two."

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+      "typ": "COS"
+    },
+    "signature":"hiyDh... (Cosigner's signature)"
+  }
+]}
+```
+
+### Protected Header Claims
+
+Protected headers in JWS [(RFC-7515 Section 2)](https://datatracker.ietf.org/doc/html/rfc7515#section-2) are used in JWS as a place to put signature header parameters. That is metadata about the signature necessary to verify the signature. For instance parameters such as, `alg` which specifies the algorithm used to generate the signature or `kid` which is the Key ID of the public key used to verify the signature.  In PK Tokens we use protected headers in a new way namely to store claims the signature is making about the identity. Next, we will look at why we need to use protected headers in this new way.
```
asamborski, line 163: Suggestion: simplify to "That is metadata necessary to verify the signature."

asamborski: Wording: "For instance, parameters such as, `alg` specifies the algorithm used to generate the signature amd `kid`, which is the Key ID of the public key used to verify the signature, may be included as metadata.

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+In an ID Token or JWT, the claims the issuer is making are set in the payload. However this isn't sufficient for OpenPubkey, signing parties other than the issuer may wish to add additional claims along with their signature. For instance a signing party who independently authenticated the identity, might want to add an additional claim specifying the time at which this independent authentication took place. This party can not update the payload without breaking the OP's signature on the payload.Our solution is to allow signing parties to specify additional claims in the protected header that signing party creates. This enables the signing party to add claims without requiring any new signatures from any other parties.
```
asamborski, line 165: Missing space between "payload.Our"

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+#### Required Claims
+
+In PK Tokens the following claims are required and assumed to exist `alg`, `typ` and `kid`. The one exception is that since OP (OpenID Providers) operate outside the OpenPubkey protocol they can generate an OP signature and OP protected header that may not include `typ` or `kid`. However in all other signatures in a PK Token, OpenPubkey can enforce that `alg`, `typ` and `kid` are specified.
```
asamborski, line 169: I'm not sure I'm following what we mean here. If OPK is built off of existing compatibility with OIDC providers, how can we assume that `typ` and `kid` exist if OIDC providers operate outside of that sphere? Do we have additional detail here on which OPs don't support these claims and what they use instead?

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+#### Custom Claims
+
+OpenPubkey allows third party to extend OpenPubkey and specify any custom claims in the protected header of PK Token signatures. The only rule is that custom claims can not use the keys: `alg`, `typ` and `kid`.
```
asamborski, line 173: *third parties

docs/pktoken.md
EthanHeilman, line 3:
```suggestion
OpenPubkey works by binding a public key to an identity described in an ID Token. To provide verifiers with all the needed information to verify this binding between identity and public key we include additional information by extending the ID Token with additional signatures. This is possible because ID Tokens are JSON Web Signatures (JWS) and can support multiple signatures. We call an ID Token extended in this fashion a PK Token (Public Key Token).
```

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+### Signature Type (typ)
+
+We use the `typ` value in the protected header of each signature to distinguish the "type" of signature it is. This is already an established pattern with OpenID Provider signatures in ID Tokens having `typ=JWT`.
+As shown, we have three signatures types:
+
+1. `typ=JWT` **OP (OpenID Provider) signature and protected header:** The first signature is the signature of the party that issued the ID Token, that is, the signature of the OpenID provider.
+2. `typ=CIC` **CIC (Client-Instance Claims) signature and protected header:** The second signature is generated by the identity's client. This signature contains user's public key.
+3. `typ=COS` **COS (Cosigner) signature and protected header:** The third signature is the COS (Cosigner) signature. The Cosigner is a third party who has independently authenticated the identity. It exists to remove the OpenID Provider as a single point of compromise. The COS signature is optional and not every PK Token will have one. It is up to OpenPubkey verifiers to decide if they require a Cosigner signature or not.
+
+### Commitment mechanism
```
EthanHeilman, line 186:
```suggestion
### Commitment Mechanism
```

docs/pktoken.md
EthanHeilman, line 14:
```suggestion
    - [Commitment Mechanism](#commitment-mechanism)
```

docs/pktoken.md
EthanHeilman, line 38:
```suggestion
  - [Our JWS Conventions](#our-jws-conventions)
```

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+```
+
+# Notes
+
+## GQ Signatures
+
+
+GQ (Guillou and Quisquater) signatures were invented in the paper ['A “Paradoxical” Indentity-Based Signature Scheme Resulting from Zero-Knowledge' (1988).](https://link.springer.com/content/pdf/10.1007/0-387-34799-2_16.pdf) GQ signatures were standardized in  GQ1 in [ISO/IEC 14888-2:2008](https://www.iso.org/standard/44227.html). Our GQ signatures are based on this standard but we have increased the security parameter to 256-bits.
+
+## Our JWS conventions
```
EthanHeilman, line 661:
```suggestion
## Our JWS Conventions
```

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+    - [GQ-Commitment PK Tokens (Gitlab-CI Example)](#gq-commitment-pk-tokens-gitlab-ci-example)
+      - [GQ-Commitment to the CIC](#gq-commitment-to-the-cic)
+      - [Security](#security)
+    - [ZK PK Tokens (Under development)](#zk-pk-tokens-under-development)
+  - [PK Token Compact Serialization](#pk-token-compact-serialization)
+    - [Refreshed Payload and Signature](#refreshed-payload-and-signature)
+- [Notes](#notes)
+  - [GQ Signatures](#gq-signatures-1)
+  - [Our JWS conventions](#our-jws-conventions)
+    - [The TYP pattern](#the-typ-pattern)
```
EthanHeilman, line 39:
```suggestion
    - [The TYP Pattern](#the-typ-pattern)
```

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+## GQ Signatures
+
+
+GQ (Guillou and Quisquater) signatures were invented in the paper ['A “Paradoxical” Indentity-Based Signature Scheme Resulting from Zero-Knowledge' (1988).](https://link.springer.com/content/pdf/10.1007/0-387-34799-2_16.pdf) GQ signatures were standardized in  GQ1 in [ISO/IEC 14888-2:2008](https://www.iso.org/standard/44227.html). Our GQ signatures are based on this standard but we have increased the security parameter to 256-bits.
+
+## Our JWS conventions
+
+In this section we discuss how the patterns we use for PK Tokens can be more broadly used for JSON Web Signatures in general.
+
+### The TYP pattern
```
EthanHeilman, line 665:
```suggestion
### The TYP Pattern
```

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+      - [GQ-Commitment to the CIC](#gq-commitment-to-the-cic)
+      - [Security](#security)
+    - [ZK PK Tokens (Under development)](#zk-pk-tokens-under-development)
+  - [PK Token Compact Serialization](#pk-token-compact-serialization)
+    - [Refreshed Payload and Signature](#refreshed-payload-and-signature)
+- [Notes](#notes)
+  - [GQ Signatures](#gq-signatures-1)
+  - [Our JWS conventions](#our-jws-conventions)
+    - [The TYP pattern](#the-typ-pattern)
+    - [Protected header claims](#protected-header-claims-1)
```
EthanHeilman, line 40:
```suggestion
    - [Protected Header Claims](#protected-header-claims-1)
```

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+One approach is to use signature order. For instance, we could specify that the first signature in the signatures list is the party that created the payload, the second signature is the cosigner, and so on. This approach has two major drawbacks. First, the order of the signatures is not signed or enforced in anyway. This means we can not assume that software and libraries won't reorder the signature list, breaking our ability to match signers to signatures. Second, this approach doesn't solve the problem of optional signers, who may or may not be required.
+
+Instead we use the `typ` (type) key in the protected header to specify the different types of signatures. For instance if the protected header of a signature has `typ=COS` it is a cosigner signature. In JWT's it is customary to have a `typ=JWT`. We extend this so that signatures each signing party in a protocol specifies their role in the `typ` key of their protected header.
+
+### Protected header claims
```
EthanHeilman, line 673:
```suggestion
### Protected Header Claims
```

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+In a JWT the claims the issuer is making are set in the payload. However this isn't sufficient for OpenPubkey, parties may wish to add additional claims although with their signature. For instance a signing party who independently authenticated the identity, might want to add an additional claim specifying the time at which this independent authentication took place. This signature can not update the payload without breaking the signature that already signed the payload.
+
+Our solution is to allow signing parties to specify additional claims in their protected header. This enables the party to add claims without requiring any new signatures from any other parties and also makes it clear who added these claims.
+
+### Compact representations of a JWS with two or more signatures
```
EthanHeilman, line 679:
```suggestion
### Compact Representations of a JWS with Two or More Signatures
```

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+      - [Security](#security)
+    - [ZK PK Tokens (Under development)](#zk-pk-tokens-under-development)
+  - [PK Token Compact Serialization](#pk-token-compact-serialization)
+    - [Refreshed Payload and Signature](#refreshed-payload-and-signature)
+- [Notes](#notes)
+  - [GQ Signatures](#gq-signatures-1)
+  - [Our JWS conventions](#our-jws-conventions)
+    - [The TYP pattern](#the-typ-pattern)
+    - [Protected header claims](#protected-header-claims-1)
+    - [Compact representations of a JWS with two or more signatures](#compact-representations-of-a-jws-with-two-or-more-signatures)
```
EthanHeilman, line 41:
```suggestion
    - [Compact Representations of a JWS with Two or More Signatures](#compact-representations-of-a-jws-with-two-or-more-signatures)
```

docs/pktoken.md
EthanHeilman, line 72:
```suggestion
[JWT (JSON Web Token)](https://datatracker.ietf.org/doc/html/rfc7519) is a type of JWS used by one party, the issuer, to make claims about another party, the subject. The issuer includes their identity in the JWT using the `iss` claim. JWTs are defined as having only one signature, the signature of the issuer.
```

docs/pktoken.md
EthanHeilman, line 111:
```suggestion
Now let's look at what happens when we extend this ID Token with additional signatures to create a PK Token.
```

docs/pktoken.md
EthanHeilman, line 163:
```suggestion
Protected headers in JWS [(RFC-7515 Section 2)](https://datatracker.ietf.org/doc/html/rfc7515#section-2) are used in JWS as a place to put signature header parameters. These parameters are metadata necessary to verify the signature. For instance, parameters such as, alg specifies the algorithm used to generate the signature amd kid, which is the Key ID of the public key used to verify the signature, may be included in the protected header.  In PK Tokens we use protected headers in a new way namely to store claims the signature is making about the identity. Next, we will look at why we need to use protected headers in this new way.
```

docs/pktoken.md
EthanHeilman, line 165:
```suggestion
In an ID Token or JWT, the claims the issuer is making are set in the payload. However this isn't sufficient for OpenPubkey, signing parties other than the issuer may wish to add additional claims along with their signature. For instance a signing party who independently authenticated the identity, might want to add an additional claim specifying the time at which this independent authentication took place. This party can not update the payload without breaking the OP's signature on the payload. Our solution is to allow signing parties to specify additional claims in the protected header that signing party creates. This enables the signing party to add claims without requiring any new signatures from any other parties.
```

docs/pktoken.md
EthanHeilman, line 169:
```suggestion
In PK Tokens the following claims are required and assumed to exist `alg`, `typ` and `kid` for all OpenPubkey generated signatures except OP (OpenID Providers) generated signatures.

OP (OpenID Providers) operate outside the OpenPubkey protocol we can not assume all OPs will generate an OP signature and OP protected header that will define a `typ` or `kid`. OP signatures are required by OpenID Connect to specify an `alg`. For OP signatures we have special logic to account for cases in which a `typ` or `kid` is not set. In all other signatures in a PK Token, OpenPubkey can enforce that `alg`, `typ` and `kid` are specified.
```

docs/pktoken.md
EthanHeilman, line 173:
```suggestion
OpenPubkey allows third parties to extend OpenPubkey and specify any custom claims in the protected header of PK Token signatures. The only rule is that custom claims can not use the keys: `alg`, `typ` and `kid`.
```

docs/pktoken.md
EthanHeilman: We can't assume they exist, so we need to be careful and handle the case where the OP doesn't. Let me reword.

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+For instance Docker uses the custom claim `att` in the CIC protected header to ensure [a particular PK Token can only be used to verify a particular object.](https://github.com/openpubkey/openpubkey/issues/33)
+
+### Signature Type (typ)
+
+We use the `typ` value in the protected header of each signature to distinguish the "type" of signature it is. This is already an established pattern with OpenID Provider signatures in ID Tokens having `typ=JWT`.
+As shown, we have three signatures types:
+
+1. `typ=JWT` **OP (OpenID Provider) signature and protected header:** The first signature is the signature of the party that issued the ID Token, that is, the signature of the OpenID provider.
+2. `typ=CIC` **CIC (Client-Instance Claims) signature and protected header:** The second signature is generated by the identity's client. This signature contains user's public key.
```
asamborski, line 183: Nit: "the user's" instead of only "user's"

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+3. `typ=COS` **COS (Cosigner) signature and protected header:** The third signature is the COS (Cosigner) signature. The Cosigner is a third party who has independently authenticated the identity. It exists to remove the OpenID Provider as a single point of compromise. The COS signature is optional and not every PK Token will have one. It is up to OpenPubkey verifiers to decide if they require a Cosigner signature or not.
```
asamborski, line 184: Small comment - some places we have OpenID Provider and others OpenID provider. May consider standardizing across one format

docs/pktoken.md
asamborski, line 186: Consider capitalizing M in mechanism for consistency across headers

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+| GQ-Commitment                 | `OP-protected header`| `CIC`        |
+| ZK-Commitment                 | `OP-protected header`| `CIC`        |
+
+### Types of Signatures in a PK Token
+
+#### OP (OpenID Provider) Signature
+
+This is signature of the OpenID Provider that created the payload and signed it to create the ID Token. This signature is responsible for binding the identity claims in the payload to that identity's public key.
+
+Typically OP signature/protected header along with the payload is the ID Token. For some more advanced forms of PK Tokens such as GQ or ZK PK Tokens, we transform OP signature and protected header. In these cases, you can no longer recover the ID Token from the PK Token, but even when we alter the OP Signature and protected header in this way, we don't break the cryptographic binding between payload and OpenID Provider that the signature provides.
```
asamborski, line 206: Similar feedback - sometimes we have OP signature and others OP Signature. Suggest sticking to one format for consistency

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+Some OpenID Providers do not specify `typ`. To get around this we classify a signature as being from the OpenID Provider if either `typ=JWT` or `typ` is not defined.
+
+#### CIC (Client-Instance Claims) Signature
+
+The Client-Instance is the identity's OpenPubkey client. The CIC are the claims made about the identity by this client. The CIC signature performs two functions. First, it provides the needed data to allow verifiers to check that an ID Token has a binding to the user's public key by opening the commitment. Second, it functions as a [Proof-of-Possession](https://csrc.nist.gov/glossary/term/proof_of_possession) showing that the identity's knows the signing key associated with the public key `upk`. This Proof-of-Possession prevents rogue key attacks in which a attackers associates their identity with another identity's public key and the attempts to claim that they produced signatures produced by the other identity.
```
asamborski, line 212: Spell check: "in which **an attacker** associates their identity with another identity's public key and the**n** attempts to claim that they produced signatures **from** the other identity." (modifications in bold)

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+The CIC always contains the following claims (and may contain other claims):
+
+* `alg` - the algorithm of both the signature and the identity's public key.
+* `upk` - the JWK (JSON Web Key) of the identity's public key. UPK stands for User's Public Key.
+* `typ` - this value is always set to `CIC` to identify this protected header and signature pair as the CIC.
+
+#### COS (Cosigner) Signature
+
+The Cosigner is a third party who issues a signature if they are able to authenticate the identity in the ID Token. This authentication must be independent of the OpenID Provider's authentication. The purpose of the Cosigner is enable OpenPubkey to maintain security even if the OpenID Provider becomes maliciously.
```
asamborski, line 222: Spell check: "becomes **malicious**."

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+Cosigning is an optional feature of OpenPubkey and OpenPubkey can be used without a cosigner. In such cases there is no COS signature.
```
asamborski, line 224: Can remove "and OpenPubkey can be used without a cosigner." as the first part of the sentence addresses that.

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+In this section we use actual PK Tokens from to illustrate the types of PK Tokens.
+
+OpenPubkey has a number of different types of PK Tokens. A full list includes:
+
+| PK Token Type                 | OP Support               | Identity type |
+| -------------                 |--------------            | ------------- |
+| Nonce-Commitment              | Google, Azure, Okta      | Human         |
+| GQ Signed Nonce-Commitment     | Google, Azure, Okta      | Human         |
+| GQ Signed Audience-Commitment | Github, GCP              | Machine       |
```
asamborski, line 246: Nit: GitHub is the official spelling.

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+| GQ-Commitment                      | Gitlab-CI                | Machine       |
```
asamborski, line 247: *GitLab

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+    "typ":"CIC",
+    "upk":{"alg":"ES256","crv":"P-256","kty":"EC" "x":"8JAMvpmdrhiKJi9A79LHPj5CPKlyztHHEkCr6tntyq8", "y":"jRqKcX8wIU24ffb5GI6z9XlePqlP1DOxvlEwvp0wC5s"}
+  }))
+```
+
+Notice that the value which is hashed to generate the `nonce` value in the payload is exactly the value given in the protected header of the CIC. Put another way, the CIC allows the verifier to open the commitment to the identity's public key in the nonce. The value `rz` is randomly chosen by the identity's OpenPubkey client, a.k.a. the client-instance, to ensure that each time a `nonce` is generated, it is always unique and random. The value `upk` is the identity's public key.
+
+#### MFA Cosigner Signature
+
+We have included in a Cosigner signature in this example to show what what one looks like. Given that the Cosigner signature does not not differ much between examples we have omitted it from the other examples. Note that PK Tokens can function without a Cosigner signature, a verifier can choose to require or not require one based on a verifiers security needs.
```
asamborski, line 328: Spell check: "a verifier**'s** security needs."

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+  },
+  "signatures": [
+    {
+      "protected": {
+      "alg": "GQ256",
+      "jkt": "w0bhEOa9d4qxGtKLGhwySJ2VZtRxPA-0abeIC9C-zPQ",
+      "kid": "eyJhbGciOiJSUzI1NiIsImtpZCI6ImUyNmQ5MTdiMWZlOGRlMTMzODJhYTdjYzlhMWQ2ZTkzMjYyZjMzZTIiLCJ0eXAiOiJKV1QifQ",
+      "typ": "JWT"
+      },
+      "signature": "hAo4...(5504 base64 characters)"
```
asamborski, line 360: Is this signature really 5000 chars long?

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+      },
+      "signature": "dsWL..."
+    }
+  ]
+}
+```
+
+#### GQ Signatures
+
+A [GQ (Guillou and Quisquater) signature](https://link.springer.com/content/pdf/10.1007/0-387-34799-2_16.pdf) is a a [Proof of Knowledge (PoK)](https://en.wikipedia.org/wiki/Proof_of_knowledge) of an RSA signature that keeps the RSA signature secret. It provides the same guarantee that the ID Token was signed by OpenID Provider, but it keeps the original RSA signature secret, preventing the ID Token in the PK Token from being used an ID Token. This works because a standard OpenID Connect-compliant service only accepts RSA signature and so will rejected an ID Token that has a GQ signature instead.
```
asamborski, line 384: Extra a in "is a Proof of Knowledge (POK)"

asamborski: *Of a RSA

asamborski: *This works because a standard OpenID Connect-compliant service only accepts RSA signature**s** and so will **reject** an ID Token that has a GQ signature instead.

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+Why do this? A non-GQ PK Token contains the ID Token signed and issued by the OP ( OpenID Provider). If the PK Token anyone who sees it could extract the ID Token including the signature issued by OP from the PK Token and then attempt to replay this ID Token to authenticate as the subject (remember ID Tokens are bearer authentication secrets). A correctly configured service would reject such a replayed ID Token because the audience (`aud`) value in the ID Token would not match the audience that the service expects. This is because ID Tokens issued for OpenPubkey will have a different audience than an ID Token issued for use in another service. Unfortunately, a common misconfiguration is that services do not check the audience claim in the ID Token. To prevent such replay attacks in OpenPubkey we use GQ signatures when a PK Token will be publicly posted.
```
asamborski, line 386: Just a suggestion: Maybe bold the "Why do this?"

asamborski: Extra space in ( OpenID Provider)

asamborski: Suggest rewording this sentence; it's a challenge to follow, "If the PK Token anyone who sees it could extract the ID Token including the signature issued by OP from the PK Token and then attempt to replay this ID Token to authenticate as the subject (remember ID Tokens are bearer authentication secrets)."

asamborski: This feels like it's worth bolding as well, "Unfortunately, a common misconfiguration is that services do not check the audience claim in the ID Token. To prevent such replay attacks in OpenPubkey we use GQ signatures when a PK Token will be publicly posted."

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+Note that in cases where a PK Token is merely used to authenticate to a server and is not made publicly available, no GQ signature is needed. For instance when OpenPubkey is used in SSH, SSH3, TLS and web authentication we do not need to use GQ signatures. When OpenPubkey is used to for software artifact signing where the signatures and PK Tokens will be posted to a public ledger, then GQ signatures or Zero Knowledge Proofs should be used.
```
asamborski, line 388: *When OpenPubkey is used **for** software artifact signing where the signatures and PK Tokens will be posted to a public ledger, then GQ signatures or Zero Knowledge Proofs should be used.

docs/pktoken.md
EthanHeilman, line 183:
```suggestion
2. `typ=CIC` **CIC (Client-Instance Claims) signature and protected header:** The second signature is generated by the identity's client. This signature's protected header contains the identity's public key.
```

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+#### GQ Signature Protected Header
+
+When we replace an RSA signature in a PK Token with a GQ signature, we also replace the protected header of the RSA signature. GQ signatures not only provide a Proof-of-Knowledge of the original RSA Signature, but they also enable anyone who knows the original RSA signature to sign a message using the RSA signature as the signing key. Using this property the GQ signature acts a signature, signing the payload and the new protected header using the original RSA signature as the signing secret.
```
asamborski, line 392: Think this should be "a RSA" but I just had a solid debate with myself about it haha! Preference is to pick one even if it's not grammatically correct and stick with it

docs/pktoken.md
EthanHeilman, line 212:
```suggestion
The Client-Instance is the identity's OpenPubkey client. The CIC are the claims made about the identity by this client. The CIC signature performs two functions. First, it provides the needed data to allow verifiers to check that an ID Token has a binding to the user's public key by opening the commitment. Second, it functions as a [Proof-of-Possession](https://csrc.nist.gov/glossary/term/proof_of_possession) showing that the identity's knows the signing key associated with the public key `upk`. This Proof-of-Possession prevents rogue key attacks in which an attacker associates their identity with another identity's public key and then attempts to claim that they produced signatures by the other identity.
```

docs/pktoken.md
EthanHeilman, line 222:
```suggestion
The Cosigner is a third party who issues a signature if they are able to authenticate the identity in the ID Token. This authentication must be independent of the OpenID Provider's authentication. The purpose of the Cosigner is enable OpenPubkey to maintain security even if the OpenID Provider becomes malicious.
```

docs/pktoken.md
EthanHeilman, line 224:
```suggestion
Cosigning is an optional feature of OpenPubkey. In such cases there is no COS signature.
```

docs/pktoken.md
EthanHeilman, line 246:
```suggestion
| GQ Signed Audience-Commitment | GitHub, GCP              | Machine       |
```

docs/pktoken.md
EthanHeilman, line 247:
```suggestion
| GQ-Commitment                      | GitLab-CI                | Machine       |
```

docs/pktoken.md
EthanHeilman, line 328:
```suggestion
We have included in a Cosigner signature in this example to show what what one looks like. Given that the Cosigner signature does not not differ much between examples we have omitted it from the other examples. Note that PK Tokens can function without a Cosigner signature, a verifier can choose to require or not require one based on a verifier's security needs.
```

docs/pktoken.md
EthanHeilman: GQ signatures are huge. They are basically a blinded RSA signature repeated under different challenge values 256  times.

docs/pktoken.md
EthanHeilman, line 388:
```suggestion
Note that in cases where a PK Token is merely used to authenticate to a server and is not made publicly available, no GQ signature is needed. For instance when OpenPubkey is used in SSH, SSH3, TLS and web authentication we do not need to use GQ signatures. When OpenPubkey is used for software artifact signing where the signatures and PK Tokens will be posted to a public ledger, then GQ signatures or Zero Knowledge Proofs should be used.
```

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+The required claims in a GQ signature protected header are:
+
+* **alg** - To signal that the GQ protected header is for a GQ signature we set `alg=GQ256`.
+* **kid** -  We set the `kid` (Key ID) of the new GQ protected to the Base64 encoding of the original RSA protected header `"kid": Base64({"alg":"RS256","kid":"e26d917b1fe8de13382aa7cc9a1d6e93262f33e2","typ":"JWT"})`. This is required because verifying the GQ signature requires the original protected header of the RSA signature.
+* **jtk** - To ensure we can always lookup and find the correct public key to verify a GQ signed ID Token, we record the `jkt` (JSON Key Thumbprint) of the OP's public key in the GQ signature's protected header. The original OP's public key for the RSA signature is needed to verify the GQ signature. This is needed because OP (OpenID Provider) are not required to set a `kid` (Key ID) or use a unique `kid` for the public keys they in their JWKS. While it is extremely rare for an OP to not use an kid or to recycle a kid, such behavior is standards compliant and we must ensure we can handle this behavior and uniquely identify the OP public key used to verify the signature.
```
asamborski, line 398: *"This is needed because OP (OpenID Provider) **is** not required to set a `kid` (Key ID) or use a unique `kid` for the public keys they **use** in their JWKS."

asamborski: *"While it is extremely rare for an OP to not use **a** kid"

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+    "upk": {
+        "alg": "ES256",
+        "crv": "P-256",
+        "kty": "EC",
+        "x": "5BP8B8bXgf0OFxHLJS5LSFlPOsfdIvf2tJU_3mwTGNE",
+        "y": "7KzWJi88qdZOI_j-kUG2aPjkzEA7IGMXFp1f-jdt28I"
+    }}))
+```
+
+User identity ID Token issuance flows set the `aud` to a unique identifier to scope the ID Token to a particular service or OIDC client. This is done, among other reasons to prevent a malicious service from replaying the ID Tokens it receives from users to impersonate those users to a another service. Allowing the requesting party to specify the `aud` as is done in machine identity would be insecure for user identity. However it is both secure and the primary pattern in machine identity flows.
```
asamborski, line 507: Extra 'a' in here: "to impersonate those users to another service."

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+#### GQ Signatures Are Required For Aud-Commitment PK Tokens
+
+The main use case of machine identity audience-commitment PK Tokens is to create publicly published signatures. As in such cases GQ signatures or ZK proofs should always be used, in such cases we require the use of GQ signatures. This eliminates the risks of accidental misconfigurations where a GQ signature should be used but is not used.
```
asamborski, line 511: Suggest rewording to remove the repetition: "As in such cases GQ signatures or ZK proofs should always be used, in such cases we require the use of GQ signatures." to "In such cases, GQ signatures or ZK proofs should always be used."

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+### GQ-Commitment PK Tokens (Gitlab-CI Example)
```
asamborski, line 513: *GitLab

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+GQ-commitment PK Tokens are designed for the case where a OP can not support a nonce-commitment or an audience-commitment. Instead the GQ signature itself functions as the binding between the identity's public key (and CIC protected header) and the ID Token. So far we have only encountered one OP which can't support a nonce-commitment or an aud-commitment: gitlab-CI.
```
asamborski, line 515: *an OP

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+**Critical:** GQ-commitment PK Tokens should only be used if an OP can not support nonce or audience-commitment PK Tokens. Never verify a GQ-bound PK Token unless the issuing OP only supports GQ-bound PK Tokens.
```
asamborski, line 517: Is this an easy way to catch malicious PK Tokens (if you see a GQ-bound PK Token from an OP that does not only support GQ-bound PK Tokens?)

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+        "y": "pfsH8--s5c8u4DxXto0sN4g5n6SjlXn1WjzaKXrr9b4"
+    }
+  },
+  "signature":"xpt6tTEzAqPrXCzAL6UCdiIliPPejRmn2sW1_RSxCt3WUQN38s3z_MURF3Bd8mlncU1UhWwSMCAcwDijm-Hc6w"}
+]}
+```
+
+#### GQ-Commitment to the CIC
+
+For Google and github we bind the CIC (Client Instance Claims) that contains the user's public key to the ID Token by putting a hashed commitment to the CIC in one of the claims of the ID Token. For Google we put the commitment in the `nonce` claim, for github we put the commitment in the audience (`aud`) claim. Gitlab-CI does not provide an `nonce` claim and does not allow a running job/workflow to specify the audience. While Gitlab does allow customizing the audience, this custom audience is a fixed configuration value and can not be set per request. As such we can not use audience-commitment PK Tokens for gitlab-CI. As gitlab-CI is for machine identity it does not support specifying a nonce and thus nonce-commitment PK Tokens are not available either.
```
asamborski, line 580: *GitHub

asamborski: *GitLab

asamborski: * a `nonce` claim

docs/pktoken.md
EthanHeilman: Fixed

docs/pktoken.md
EthanHeilman: Fixed

docs/pktoken.md
EthanHeilman: I went with "an RSA"

docs/pktoken.md
EthanHeilman: Fixed

docs/pktoken.md
EthanHeilman: Fixed

docs/pktoken.md
EthanHeilman: Fixed

docs/pktoken.md
EthanHeilman: Fixed

docs/pktoken.md
EthanHeilman: Fixed

docs/pktoken.md
EthanHeilman: An attacker is unlikely to do this because we will always reject it. It is a good check for misconfigurations

docs/pktoken.md
EthanHeilman: Fixed

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+GQ-binding solves this problem posed by gitlab-CI. Remember that with GQ signatures, we use the OP's RSA signature on the ID Token to generate a GQ signature that signs both the payload and the GQ protected header. Therefore, we can simply put the commitment to the identity's public key, i.e. the hash of the CIC, in the GQ protected header and then when the GQ signature will also sign this commitment. GQ-bound PK Tokens were introduced in PR: [Adds gitlab-ci OP using GQ commitment binding](https://github.com/openpubkey/openpubkey/pull/143).
```
asamborski, line 584: Not sure I'm following this last part of the sentence here "and then when the GQ signature will also sign this commitment." Do you mean we put the CIC hash in both places?

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+        "x": "c63goURlnP5vbJbt4chtOHTHwg6Yvy4h6_aw3Zc2A5o",
+        "y": "pfsH8--s5c8u4DxXto0sN4g5n6SjlXn1WjzaKXrr9b4"
+    }))
+```
+
+and the `aud` is always prefixed with `"OPENPUBKEY-PKTOKEN:"`. In the next section we explain why we require the audience to be prefixed this way.
+
+#### Security
+
+If configured correctly, GQ-bound PK Tokens offer the same security as nonce or audience bound PK Tokens. However GQ-commitment PK Tokens do introduce a security risk from misconfiguration not present in the other cases. The risk is that if a GQ signature is not used and an attacker learns the ID Token, the attacker can compute the GQ signature from the RSA signature on a CIC protected header of the attackers choosing.
```
asamborski, line 606: *attacker**'s** choosing

docs/pktoken.md
EthanHeilman: I think I got all of them

docs/pktoken.md
EthanHeilman: Tried the bolding and rewording

docs/pktoken.md
```diff
@@ -0,0 +1,707 @@
...
+        "y": "pfsH8--s5c8u4DxXto0sN4g5n6SjlXn1WjzaKXrr9b4"
+    }
+  },
+  "signature":"xpt6tTEzAqPrXCzAL6UCdiIliPPejRmn2sW1_RSxCt3WUQN38s3z_MURF3Bd8mlncU1UhWwSMCAcwDijm-Hc6w"}
+]}
+```
+
+#### GQ-Commitment to the CIC
+
+For Google and github we bind the CIC (Client Instance Claims) that contains the user's public key to the ID Token by putting a hashed commitment to the CIC in one of the claims of the ID Token. For Google we put the commitment in the `nonce` claim, for github we put the commitment in the audience (`aud`) claim. Gitlab-CI does not provide an `nonce` claim and does not allow a running job/workflow to specify the audience. While Gitlab does allow customizing the audience, this custom audience is a fixed configuration value and can not be set per request. As such we can not use audience-commitment PK Tokens for GitLab-CI. As GitLab-CI is for machine identity it does not support specifying a nonce and thus nonce-commitment PK Tokens are not available either.
```
EthanHeilman, line 584:
```suggestion
For Google and github we bind the CIC (Client Instance Claims) that contains the user's public key to the ID Token by putting a hashed commitment to the CIC in one of the claims of the ID Token. For Google we put the commitment in the `nonce` claim, for github we put the commitment in the audience (`aud`) claim. Gitlab-CI does not provide a `nonce` claim and does not allow a running job/workflow to specify the audience. While Gitlab does allow customizing the audience, this custom audience is a fixed configuration value and can not be set per request. As such we can not use audience-commitment PK Tokens for GitLab-CI. As GitLab-CI is for machine identity it does not support specifying a nonce and thus nonce-commitment PK Tokens are not available either.
```

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+## Our JWS conventions
+
+In this section we discuss how the patterns we use for PK Tokens can be more broadly used for JSON Web Signatures in general.
+
+### The TYP pattern
+
+Because a PK Token always has more two or more signatures we have been forced to think about how to effective organize a JWS (JSON Web Signatures) that two or more signatures. If a JWS has only one signature, then the `iss` (issuer) claim in the payload identifies the party that generated the signature. Once you have two or more signatures, how do you determine which signature matches the issuer? How do you determine the identity of signer that generated each signature? Determining what party generated what signature is important because we need to know where to find which public key to verify which signature.
```
asamborski, line 672: *"how to effective**ly** organize"

asamborski: *"that **is** two or more signatures."

asamborski: This sentence is a bit confusing. Is there a way to reword it to make it easier to parse? "Determining what party generated what signature is important because we need to know where to find which public key to verify which signature."

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+
+Instead we use the `typ` (type) key in the protected header to specify the different types of signatures. For instance if the protected header of a signature has `typ=COS` it is a cosigner signature. In JWT's it is customary to have a `typ=JWT`. We extend this so that signatures each signing party in a protocol specifies their role in the `typ` key of their protected header.
+
+### Protected header claims
+
+In a JWT the claims the issuer is making are set in the payload. However this isn't sufficient for OpenPubkey, parties may wish to add additional claims although with their signature. For instance a signing party who independently authenticated the identity, might want to add an additional claim specifying the time at which this independent authentication took place. This signature can not update the payload without breaking the signature that already signed the payload.
```
asamborski, line 680: I think you meant to say "along with their signature" instead of "although"

docs/pktoken.md
```diff
@@ -0,0 +1,702 @@
...
+BASE64URL(UTF8(JWS Protected Header-1)) || ':' ||
+BASE64URL(UTF8(JWS Signature-1)) || ':' ||
+BASE64URL(UTF8(JWS Protected Header-2)) || ':' ||
+BASE64URL(UTF8(JWS Signature-2)) || ':' ||
+...
+BASE64URL(UTF8(JWS Protected Header-N)) || ':' ||
+BASE64URL(UTF8(JWS Signature-N))
+```
+
+We use `:` as a delimiter rather than `.` to avoid confusion arising from a someone attempting to parse this as a standard single signature JWS.
```
asamborski, line 707: Extraneous "a" in "from a someone"

docs/pktoken.md
EthanHeilman: Fixed

docs/pktoken.md
EthanHeilman: Removed the confusing sentence, fixed the other issues

docs/pktoken.md
EthanHeilman: fixed

docs/pktoken.md
EthanHeilman: Fixed

docs/pktoken.md
EthanHeilman: Rewrote

docs/pktoken.md
EthanHeilman: Fixed

PR #221, opened 2025-01-21
Adds Azure/Microsoft provider

### LINKED ISSUE #6: Azure OIDC provider

EthanHeilman:
This PR Fixes #6 by adding an Azure/Microsoft OP (OpenID Provider) and a unittest.

- It also refactors out shared code between Google and Azure into a standard provider and a util file
- Creates an example project for Azure
- Fixes minor bug in Google OP where a refresh token was not being issued

## TODOs

- [X] Pull all of the shared code between the Google and Azure provider into a standard provider file
- [X] Move test to the standard provider

## Notes 
I had originally planned to use the Azure provider in OPK-SSH as part of open sourcing that code, but there doesn't appear to be an Azure provider in OPK-SSH so I just wrote the Azure provider from scratch.

## Tests

- Tested manually on windows on brave without an MFA cosigner

As PR #220 builds on this and will require testing on all major operating systems and browsers. I am delaying testing to PR #220

## Video

https://github.com/user-attachments/assets/f44dcd05-2c98-49bf-933c-17ee175f1890


providers/azure.go
```diff
@@ -0,0 +1,128 @@
...
+	// More details can be found at
+	// https://learn.microsoft.com/en-us/entra/identity-platform/access-tokens
+	TenantID string
+}
+
+func GetDefaultAzureOpOptions() *AzureOptions {
+	defaultTenantID := "9188040d-6c67-4c5b-b112-36a304b66dad"
+	return &AzureOptions{
+		Issuer:   azureIssuer(defaultTenantID),
+		ClientID: "bd345b9c-6902-400d-9e18-45abdf0f698f", // TODO: replace with a better client ID
```
jonnystoten, line 75: I wonder if we should really have a default client ID :thinking: Perhaps this should be required to be set by the calling code, WDYT? (Probably not for this PR as the google one also has a hard-coded default one.)

examples/azure/example.go
```diff
@@ -0,0 +1,200 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Pretty print our json token
+	var prettyJSON bytes.Buffer
+	if err := json.Indent(&prettyJSON, pktJson, "", "  "); err != nil {
+		return err
+	}
+	fmt.Println(prettyJSON.String())
```
jonnystoten, line 158:
```suggestion
	pktJson, err := json.MarshalIndent(pkt, "", "  ")
	if err != nil {
		return err
	}

	fmt.Println(string(pktJson))
```

providers/utils.go
```diff
@@ -0,0 +1,68 @@
...
+func FindAvailablePort(redirectURIs []string) (*url.URL, net.Listener, error) {
+	var ln net.Listener
+	var lnErr error
+	for _, v := range redirectURIs {
+		redirectURI, err := url.Parse(v)
+		if err != nil {
+			return nil, nil, fmt.Errorf("malformed redirectURI specified, redirectURI was %s", v)
+		}
+
+		lnStr := fmt.Sprintf("localhost:%s", redirectURI.Port())
```
jonnystoten, line 39: Hmm, do we really want to just take the port and change the host to localhost? Perhaps we should instead be skipping any redirect URIs that don't have a `localhost` host.

jonnystoten:
> Pull all of the shared code between the Google and Azure provider into a standard provider file

Very nice!

providers/azure.go
EthanHeilman: I want someone to just be able to download OpenPubkey and use it out of the box. Without a default client ID they would first need to sign into Azure or GCP and go through that whole process and then configure the clientID in the provider.

With OPKSSH we should be able to be able to use a configure file for this.

providers/utils.go
EthanHeilman: Good point, added a check

Issue #276, opened 2025-02-28
Use an unprivileged user on the SSH server for opkssh verify


Lekensteyn:
In the [README](https://github.com/openpubkey/openpubkey/blob/main/opkssh/README.md) and the [installation script](https://github.com/openpubkey/openpubkey/blob/c28020447a4f6cdaeee96c42702fa57c305580db/opkssh/scripts/install-linux.sh#L71), you run the tool as root user. This opens up an attack vector and contradicts the recommendation "to use a dedicated user that has no other role on the host than running authorized keys commands".

It appears that the SSH server only needs:
- The ability to execute the `opkssh verify` command. Typically in `/usr/local/bin/opkssh`, it does not not belong in `/etc/` (#271)
- `opkssh verify` needs read-only access to `/etc/ssh/auth_id` and`/etc/ssh/providers`

Proposed solution:
- Create a new dedicated system user and group, `opkssh`
- Restrict file permissions such that the `opkssh` group has readonly access.
- Adjust the SSHD config to use the new `opkssh` user instead of `root`.

These instructions should be sufficient to achieve that (minus the `sshd_config` change):
```sh
useradd --system --shell /usr/sbin/nologin --no-create-home --homedir /nonexistent opkssh
install -dm755 /etc/opk
install -m640 -g opkssh /dev/null /etc/opkssh/providers
install -m640 -g opkssh /dev/null /etc/opkssh/auth_id
````

EthanHeilman:
I spent today looking [into using a less permissioned user for the AuthorizedKeysCommand](https://github.com/openpubkey/openpubkey/pull/280). The solution is more complex than it at first appears.

The AuthorizedKeysCommand has to read a file in the user's home directory, e.g. `/home/alice/.opk/auth_id`. This means the AuthorizedKeysCommand has to transverse /home, /home/alice/, /home/alice/.opk so I would need add +x to `/home/alice` which would require changing the file permissions on `/home/alice` which seems unacceptable to do to someones system.

I could grant the unprivileged user a sudoers privilege of being able to run stat everywhere, which if can let me get past the directories to the file but also seems over-privileged.

Probably the most secure approach would be to create a simple script which can only read files that match the pattern `/home/*/.opk/auth_id` and then give it sudoer stat access. I worry this may cause compatibility issues down the road because of all the moving peices, but seems doable and I'm prototyping it now.

[...]

PR #280, opened 2025-03-01
Uses less privileged user for SSH AuthorizedKeysCommand user

EthanHeilman:
- Renames log file from `/var/log/openpubkey.log` to `/var/log/opkssh.log`
- Creates opksshuser to be used as less privileged AuthorizedKeyCommand user 
  - Fixes https://github.com/openpubkey/openpubkey/issues/276
- Creates gocode which can read policies in user's home directories because AuthorizedKeyCommand user is no longer root
- Gets sudoer behavior working in golang code eliminating need to additional script
  - Fixes https://github.com/openpubkey/openpubkey/issues/286
- Modifies install script so that it can install from a local binary, allowing us to delete the install command
  - Fixes https://github.com/openpubkey/openpubkey/issues/285
  
## How our less privileged AuthorizedKeyCommandUser works

OPKSSH requires that sshd run the AuthorizedKeyCommand using a specific user, a.k.a. the AuthorizedKeyCommandUser. Prior to this PR we had been using root for this. From a usability standpoint root is ideal. However from a security standpoint calling opkssh as root from sshd presents a security risk: if the opkssh was compromised, or tricked as in the case of [confused deputy attacks](https://en.wikipedia.org/wiki/Confused_deputy_problem), then an attacker might be able to assume root privileges. Put another way, we should not over-privilege the AuthorizedKeyCommand call to opkssh.

SSH recommends that AuthorizedKeyCommandUser be an unprivileged and dedicated user. This PR switches us to using such a user.

```bash
AuthorizedKeysCommand /usr/local/bin/opkssh verify %u %k %t
AuthorizedKeysCommandUser opksshuser
```

Rather than root we now create and use the linux user _opksshuser_ which is a member of the group _opksshgroup_. When opkssh is called as an AuthorizedKeysCommand by sshd it must read the following files:

- `/etc/opk/provider`,
-  `/etc/opk/auth_id` 
- `/home/{USER}/.opk/auth_id` where {USER} is the linux principal that the party sshing in is attempt to assume.

We set `/etc/opk/provider` and `/etc/opk/auth_id` to have the linux permissions `640` i.e. write/read by owner, read by group, nothing for everyone. We set the group of these two files to be opksshgroup. Since opksshuser is in the group opksshgroup it has read permissions for these files.

```bash
-rw-r----- 1 root opksshgroup   0 Mar 10 20:55 auth_id
-rw-r----- 1 root opksshgroup 310 Mar 10 20:55 providers
```

Reading the file  `/home/{USER}/.opk/auth_id` does not have such an easy solution. This is because while we can give opksshuser permission to read file `auth_id` in  `/home/{USER}/.opk/` it does not have permission to transverse the directory path `/home/{USER}/` to read the file. Changing the file permissions on a user's entire home directory is abusive and violates the common expectations. Additionally, if the user every changed the file permissions on their home directory back, we start failing.

We add a file to ` /etc/sudoers.d/` which contains:

```
opkssh ALL=(ALL) NOPASSWD: ALL=(ALL) NOPASSWD: /usr/local/bin/opkssh readhome *
```

The command `opkssh readhome <username>` outputs the contents of the policy file `/home/<username>/opk/.auth_id` policy file in the user's home directory for the specified `<username>`. To enable the opkssh process which is being run by the AuthorizedKeysCommand to specify any username, we use [the sudoers wildcard](https://linux.die.net/man/5/sudoers), 
where `*` allows any string of characters to be used.

While this is a security improvement over running a root AuthorizedKeysCommandUser, it does have some risks which we protect.

**Symlinks** First, users could put a symlinks in their home directory named `auth_id` and then attempt to connect over ssh to trick sshd into reading the file pointed to by the symlinks. Such an attack doesn't allow a user to read the file pointed to by the symlinks but they may be able to infer if the file exists and how large it is based on how long sshd takes to reject the connection.

To mitigate this risk we open the file with O_NO_FOLLOW to ensure if the file is a symlink we fail on file open. This makes the symlink check atomic with the file read and prevents the TOCTOU (Time-of-Check to Time-of-Use) vulnerability discussed in the notes below.

To ensure that `opkssh readhome` can't be used to read something which the user identified by the username is not allowed to read, readhome checks the following:
- the file is not a symlink,
- the file path is `~/.opk/auth_id`
- the user is the owner of the file
- the file has permissions `0600` rwx for owner and nothing for anyone else

## TODOs

 - [x] Add opksshuser and opksshgroup group create to install scripts
 - [x] Test on fresh install
- [X] Write up PR description
- [X] Document security properties of this change
- [X] Update all documentation around how to configure a system
- [x] Delete install command
- [x] Get intregration tests passing
- [x] Rearchitecture the file loaders so that they can handle different file permissions
- [X] opkssh must check that `opkssh_read_home.sh` has the correct permissions before calling it.
   - [X] Checks permissions
   - [x] Check owner of file is root
- [x] Ensure the sudoers script  `opkssh_read_home.sh` won't follow symlinks 
- [X] Update all the documentation that talks about file permissions, e.g. chmod 600 etc...
- [x] Resolve all TODO comments in code
- [x] Ensure policy loaders are clean and doc strings

After first review TODOs:
- [x] Use opkssh/opkssh instead of opksshuser/opksshgroup as the user/group roles are implied. Both are intended to refer to the same entity, so having different names is unusual and create more confusion.
- [x] `opkssh readhome <username>` for sudoer access to home policy.
- [x] instead create a file in /etc/sudoers.d/. Ideally create a single file that can be installed in distribution packages.
- [x] See if I can get away with not using any sudo commands in the install script
- [x] Figure out way to prevent "TOCTOU (time-of-check to time-of-use) race condition. Between checking with stat and reading the contents with cat, a local attacker can change the file that is being checked."

## Notes on TOCTOU symlink issue and alternative approaches 

These notes describe some approaches I attempted before deciding on the `readhome` approach. I wanted to document them in case I come back to this problem.

### Approach: stat and then cat

**Note we did not use this approach because this approach is secure**

We run `cat` and `stat` on  `/home/{USER}/.opk/auth_id` in opkssh to read these files. To enable this we must add the following lines to `/etc/sudoers`

```bash
opksshuser ALL=(ALL) NOPASSWD: /bin/cat /home/*/.opk/auth_id
opksshuser ALL=(ALL) NOPASSWD: /bin/stat *
```

Thus, to read `/home/{USER}/.opk/auth_id`, opkssh running as opksshuser  runs `stat` against `/home/{USER}/.opk/auth_id` to determine if the permissions are safe, and if the permissions are safe, it runs `cat /home/{USER}/.opk/auth_id` so that the opkssh binary can see the contents of the file.

While this is a security improvement over running a root AuthorizedKeysCommandUser, it does have some risks which we protect.

**Symlinks** First, users could put a symlinks in their home directory named `auth_id` and then attempt to connect over ssh to trick sshd into reading the file pointed to by the symlinks. Such an attack doesn't allow a user to read the file pointed to by the symlinks but they may be able to infer if the file exists and how large it is based on how long sshd takes to reject the connection.

We have mitigated this risk by checking if the file is symlink before reading it.  However this mitigation does not work because a user could perform the following attack.

1. Make `~/.opk/auth_id` not a symlink.
2. After `stat` runs and passes, but before cat runs, they user could change `~/.opk/auth_id`  into a symlink.

**edit:** There appears to be no way to atomically check if a file is a symlink while reading the file using linux commands. It requires sending the O_NO_FOLLOW flag on file open and linux commands do not expose this. I really think they should, but no one asked me.

### Approach: cat with the user whose home directory you are reading

Adding this to the sudoer file
```
opkssh ALL=(ALL) NOPASSWD: /bin/su [a-z0-9_-]+ -c '/bin/cat /home/[a-z0-9_-]+/.opk/auth_id'
```
 
This allows opkssh to assume any user identity but only to run cat for files matching the pattern `/home/*/.opk/auth_id`. 

If opkssh is malicious this is no worse or better than:

```
opkssh ALL=(ALL) NOPASSWD: /bin/cat /home/*/.opk/auth_id
```

since opkssh could run `/bin/su root -c '/bin/cat /home/<ANY USER>/.opk/auth_id`, but that was what it could do before.

However this defeats symlink attacks because if opkssh is honest it sets the username to the username of the home directory of the user it is reading the file. Thus, the read is limited by the users permissions, symlink or not. Consider the user eve. opkssh will run `/bin/su eve -c `/bin/cat /home/eve/.opk/auth_id`. If eve makes auth_id a symlink to a file she does not have permissions to read, the read will fail because the cat command is being run as eve.

**edit:** I couldn't get this approach to work because I couldn't get regex to work so I abandoned it

### Approach: cat with the user whose home directory you are reading

 sudo -u {newuser} cat .opk/auth_id and configure the sudoers file to restrict the command to exact that string (no wildcards) and a fixed working directory (requires sudo 1.9.3 or newer, does not work with Debian buster/oldstable, only bullseye/stable and newer).


Lekensteyn:
Thank you for the detailed description! I have not looked thoroughly at the code yet, but based on the description I have some suggestions.

Nit: consider `opkssh`/`opkssh` instead of `opksshuser`/`opksshgroup` as the user/group roles are implied. Both are intended to refer to the same entity, so having different names is unusual and create more confusion.

Use of `sudo` creates an extra risk. While `cat /home/*/.opk/auth_id` should be safe (`*` does not match slashes according to [sudoers(7)](https://www.man7.org/linux/man-pages/man5/sudoers.5.html)), being able to check for arbitrary files as root user seems excessive. The current approach also has these issues:

- TOCTU (time-of-check to time-of-use) race condition. Between checking with stat and reading the contents with `cat`, a local attacker can change the file that is being checked. 
- If you are concerned about the `opksshuser` being compromised, then you should assume that it can execute `stat` on any file without further (symlink) checks given the `sudoers` config.

Some hardening suggestions/alternatives:
- Require filesystem support for ACLs (enabled by default), and use `setfacl -m u:opkssh:x /home/someuser /home/someuser/.opk` and `setfacl -m u:opkssh:r /home/someuser/.opk/auth_id` to set appropriate permissions. Alternatively, only use ACLs to set the `x` bit on the home directory, and change the group for the `.opk` directory and its contents to `opkssh` and grant it read (+descend for directories).
- Use an alternative shared directory with files for each user underneath, e.g. `/var/lib/opkssh/auth_ids/{username}` where the opkssh daemon can read any file and other system users can create files for themselves.
  * Set the owner/group to `root`/`opkssh`.
  * Change the mode such that the setgid (2000) bit is set (ensuring that newly created files have group `opkssh`), and the sticky bit (1000) is set, such that users can only delete files they own.
  * The final mode should be at least `3753`, granting the opkssh user read/descend access, and allowing the world to descend and write files. However given that other users can already get names and see the file sizes, you could also simply use `3777`.
  * When reading files, open the file descriptor, check that the owner of the file matches the user of the filename. File permissions takes care of the rest.
  * This scheme is vulnerable to a denial of service where local system users create files named after other users.
* If you want to continue to rely on `sudo`, consider `sudo -u {newuser} cat .opk/auth_id` and configure the sudoers file to restrict the command to exact that string (no wildcards) and a fixed working directory (requires sudo 1.9.3 or newer, does not work with Debian buster/oldstable, only bullseye/stable and newer).

Miscellaneous suggestions:
- Do not touch `/etc/sudoers/`, instead create a file in `/etc/sudoers.d/`. Ideally create a single file that can be installed in distribution packages.
- Be aware that the home directory is not necessarily `/home/{username}`. The Go code handles this well, but the `sudo` config does not.

EthanHeilman:
Excellent feedback, thank you. I am updating my TODO list to fix these issues

> Require filesystem support for ACLs (enabled by default), and use setfacl -m u:opkssh:x /home/someuser /home/someuser/.opk and setfacl -m u:opkssh:r /home/someuser/.opk/auth_id to set appropriate permissions.

I've been hesitant to use filesystem ACLs because I'm worried about running into filesystems that don't support them and I'm worried about someone resetting the home directory ACL and then being locked out of their account. I can see someone removing +x on `/home/` without realizing they are going to hard lock themselves out of their server, whereas most ssh users are aware that changing permissions on `.ssh` and related files is dangerous.

> Use an alternative shared directory with files for each user underneath, e.g. /var/lib/opkssh/auth_ids/{username} where the opkssh daemon can read any file and other system users can create files for themselves.

I want to preserve the `~/.ssh/authorized_keys` pattern used by SSH. Putting it in `/var/lib` would solve a lot of problems. ChatGPT also suggested this. I wouldn't look there, if I was user. I want to copy the conventions used by SSH and SSH3.

I'm not concerned about this attack. If the attacker can alter the 

> If you want to continue to rely on sudo, consider sudo -u {newuser} cat .opk/auth_id and configure the sudoers file to restrict the command to exact that string (no wildcards) and a fixed working directory (requires sudo 1.9.3 or newer, does not work with Debian buster/oldstable, only bullseye/stable and newer).

Great call. I don't know why I didn't think about doing this, but this is an easy win. Thanks!

> Be aware that the home directory is not necessarily /home/{username}. The Go code handles this well, but the sudo config does not.

Created a ticket to track this. Don't plan to fix anytime soon, but it would be a nice feature to have and worth documenting we don't have it

https://github.com/openpubkey/openpubkey/issues/287

EthanHeilman:
@Lekensteyn I'm not sure how your suggestion here would work

> If you want to continue to rely on sudo, consider sudo -u {newuser} cat .opk/auth_id and configure the sudoers file to restrict the command to exact that string (no wildcards) and a fixed working directory (requires sudo 1.9.3 or newer, does not work with Debian buster/oldstable, only bullseye/stable and newer).

Are you suggesting a sudoer line like:

```bash
opkssh ALL=(ALL) NOPASSWD: sudo -u [a-z0-9-_.]+ cat ~/.opk/auth_id
```

This might work if I can get regex in the sudoers file to work. So far I've been running into issues with sudo regex. I have a version that should support it, but it doesn't seem to work.

I am curious what happens if someone symlinks .opk/auth_id --> /dev/random. I really want to rule out symlinks to ensure no one figures some clever way of getting ssh to read `/proc` or `/dev`.

PR #304, opened 2025-03-27
client: pass crypto.PublicKey instead of crypto.Signer

Foxboron:
When implementing abstract crypto.Signers we can't rely on the jwt library to figure out the correct encryption algorith. Passing `crypto.PublicKey` is supported and does the correct thing.


EthanHeilman (approved):
Looks good

Issue #303, opened 2025-03-27
Add ability to build an OP from configuration supplied at runtime


EthanHeilman:
Relates to https://github.com/openpubkey/opkssh/pull/58

EthanHeilman:
Fixed in https://github.com/openpubkey/opkssh/pull/58

PR #306, opened 2025-03-28
Adding the possibility to use other IDPs than google, microsoft and g…

Hanarion:
Ref : https://github.com/openpubkey/opkssh/pull/60

This pull request includes significant changes to the `chooser` functionality to make it more dynamic and maintainable. The changes primarily focus on refactoring the code to support a more flexible way of handling multiple OpenID providers.

### Improvements to chooser functionality:

* [`client/choosers/chooser.tmpl`](diffhunk://#diff-e5a0ed895b3aa3a0d42161223e937b50faab148c81010dca2e5d785e49582502L22-R35): Refactored the template to dynamically generate provider buttons based on the `Providers` data structure, allowing for easier addition of new providers.

* [`client/choosers/web_chooser.go`](diffhunk://#diff-45d886d89330a9b6d96ac6301c031e0aeb53d864a5181e046eee574e93850cc1R101-R145): Added a `Provider` struct and modified the `ChooseOp` function to populate the provider data dynamically. This includes sorting provider names and handling both known and unknown providers more gracefully.

### Codebase simplification:

* [`client/choosers/web_chooser.go`](diffhunk://#diff-45d886d89330a9b6d96ac6301c031e0aeb53d864a5181e046eee574e93850cc1L204-R238): Simplified the `IssuerToName` function to handle unknown issuers more gracefully by returning the issuer domain name if it starts with "https://".

### Dependency management:

* [`client/choosers/web_chooser.go`](diffhunk://#diff-45d886d89330a9b6d96ac6301c031e0aeb53d864a5181e046eee574e93850cc1R29): Added the `sort` package to handle sorting of provider names.


Hanarion:
Relates to this issue : https://github.com/openpubkey/openpubkey/issues/303

Hanarion:
Using opkssh with code from the pull request https://github.com/openpubkey/opkssh/pull/60:
image

EthanHeilman:
This is really cool and something I've personally wanted in OpenPubkey since I created the web chooser. I need to eat some food, but when I come back I'll review the code.

Will https://github.com/openpubkey/opkssh/pull/60 need this change? I ask because it is does, we should merge this first and then do a OpenPubkey release.

Hanarion:
> Will [openpubkey/opkssh#60](https://github.com/openpubkey/opkssh/pull/60) need this change? I ask because it is does, we should merge this first and then do a OpenPubkey release.

For the code that is commited for now, it does not need to, but in the future, if we implement the OPKSSH_PROVIDERS mechanism, then yes it will be needed in order to show all the providers configured on opkssh in the web chooser

But as you saw, there is an issue with the tests and can't figure out why, i thought it was a problem on my end, but the same seems to happen for the CI/CD, the TestUnknownProviderError seems to hang indefinitively. I need to dig down what's going on but from what i seem to see, maybe it's a timeout issue on a unresponsive provider (if provider down or random valid dns without web server)

client/choosers/web_chooser.go
```diff
@@ -201,6 +229,12 @@ func IssuerToName(issuer string) (string, error) {
 	case strings.HasPrefix(issuer, "https://gitlab.com"):
 		return "gitlab", nil
 	default:
-		return "", fmt.Errorf("unknown OpenID Provider issuer: %s", issuer)
```
EthanHeilman, line 204: I believe the test is failing because the test is expecting to get this error back immediately but after these changes there is no such thing as an unknown provider so it completes successfully

```
func TestUnknownProviderError(t *testing.T) {
	googleOpOptions := providers.GetDefaultGoogleOpOptions()
	googleOpOptions.Issuer = "https://unknown-issuer.example.com"
	googleOp := providers.NewGoogleOpWithOptions(googleOpOptions)
```

https://github.com/openpubkey/openpubkey/blob/08989a372e547aa4499cb15f30c82ad61f02623e/client/choosers/web_chooser_test.go#L150

client/choosers/web_chooser.go
```diff
@@ -97,24 +98,51 @@ func (wc *WebChooser) ChooseOp(ctx context.Context) (providers.OpenIdProvider, e
...
+				})
+				continue
+			}
+			data.Providers = append(data.Providers, Provider{
+				Name:   providerName,
+				Button: "",
+			})
+
 		}
+
```
EthanHeilman, line 145:
```suggestion
```

client/choosers/web_chooser.go
Hanarion: Yeah i figured out that i didn't changed the tests after writing this comment
I just fixed the tests and i also added a HTTP check for the code 200 on the openid endpoint, as now the user can enter everything, better to check if he didn't make any mistakes before showing him the chooser and failing there.

client/choosers/web_chooser.go
```diff
@@ -72,13 +73,17 @@ func (wc *WebChooser) ChooseOp(ctx context.Context) (providers.OpenIdProvider, e
 
 	providerMap := map[string]providers.BrowserOpenIdProvider{}
 	for _, provider := range wc.OpList {
-		if providerName, err := IssuerToName(provider.Issuer()); err != nil {
-			return nil, err
-		} else {
-			if _, ok := providerMap[providerName]; ok {
-				return nil, fmt.Errorf("provider in web chooser found with duplicate issuer: %s", provider.Issuer())
+		if providers.CheckURLStatus(provider.Issuer() + "/.well-known/openid-configuration") {
```
EthanHeilman, line 76: We try to hide these details so that we can mock them in tests and the unittests don't depend on the external resources. For instance if example.com added the page `https://unknown-issuer.example.com/.well-known/openid-configuration` our tests would break.

I don't think we have to cover the unknown provider testcase anymore.

Hanarion: I think it's better to keep the check, but to remove the test then.
I the user enters a bad value, it's better to inform him explicitly that his provider is wrong

Are you agreeing with that ?

EthanHeilman: If you really want this check add a method to the providers interface like `TestReachable() error` for the standard provider you can hit the `/.well-known/openid-configuration` endpoint but people writing tests can mock it however they want as the details are hidden inside the interface.

https://github.com/openpubkey/openpubkey/blob/08989a372e547aa4499cb15f30c82ad61f02623e/providers/op.go#L32-L39

The most likely way this error will get hit is when an OpenID Provider is offline or the persons computer is offline. For instance open their laptop in a hotel and haven't auth'd the WIFI yet. The error message `provider issuer is not a valid openid issuer: ` is going to be very confusing them because the real problem is just that their laptop can't connect to the OpenID Provider but the configuration is actually correct.

Currently we don't assume any of the values about a provider as they are hidden behind the interface. In theory we could support a provider that doesn't have an standard `/.well-known/openid-configuration`.

EthanHeilman:
@Hanarion I'm going to working on this all Sunday, I'm going to try to merge this by Sunday.

client/choosers/web_chooser.go
Hanarion: After thinking, i don't think this check is really useful after all, as the error that will happen if there is no check is more explicit than the check message :
`Error logging in: error requesting OIDC tokens from OpenID Provider: error creating provider: Get "https://custom-idp1.tld/application/o/opkssh/.well-known/openid-configuration": dial tcp: lookup custom-idp1.tld: no such host`

providers/utils.go
```diff
@@ -19,12 +19,11 @@ package providers
 import (
 	"crypto/rand"
 	"fmt"
+	httphelper "github.com/zitadel/oidc/v3/pkg/http"
```
EthanHeilman, line 22:
```suggestion
```

providers/utils.go
```diff
@@ -19,12 +19,11 @@ package providers
...
 	"io"
 	"net"
 	"net/url"
 	"strings"
```
EthanHeilman, line 26:
```suggestion
	"strings"
	
	httphelper "github.com/zitadel/oidc/v3/pkg/http"
```

providers/utils.go
```diff
@@ -23,7 +23,7 @@ import (
...
-
+
```
EthanHeilman, line 26:
```suggestion

```

EthanHeilman (approved):
LGTM, I'm going to do a manual test of this later today using your opkssh custom provider PR. Thanks for the work on this

EthanHeilman:
@Hanarion I advertised your work on bsky. I didn't know if you had a bsky account. If you do I'll link to it https://bsky.app/profile/ethanheilman.bsky.social/post/3llost35vm22g

PR #310, opened 2025-04-08
Switch to prompt=select_account

kristiangronas:
This is already what the comment says anyway, and prompt=consent does not work in organizations where consent is restricted to admins

### DIFF
--- a/providers/standard_provider.go
+++ b/providers/standard_provider.go
@@ -115,7 +115,7 @@ func (s *StandardOp) requestTokens(ctx context.Context, cicHash string) (*simple
 		// Results in better UX than just automatically dropping them into their
 		// only signed in account.
 		// See prompt parameter in OIDC spec https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
-		rp.WithPromptURLParam("consent"),
+		rp.WithPromptURLParam("select_account"),
 		rp.WithURLParam("access_type", "offline")),
 	)
 


EthanHeilman:
Thanks for this. Originally this was:

```golang
mux.Handle("/login", rp.AuthURLHandler(state, relyingParty,
    rp.WithURLParam("nonce", cicHash),
    // Select account requires that the user click the account they want to use.
    // Results in better UX than just automatically dropping them into their
    // only signed in account.
    // See prompt parameter in OIDC spec https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest
    rp.WithPromptURLParam("select_account"),
    rp.WithPromptURLParam("consent"), // This ensures we always get a refreshed ID Token
    rp.WithURLParam("access_type", "offline")),
)
```
https://github.com/openpubkey/openpubkey/blame/178adc9b28f897c6ffa8f5ce2105d0c314308e8c/providers/standard_provider.go#L111-L120

Based on this comment and the statement ["Fixes minor bug in Google OP where a refresh token was not being issued"](https://github.com/openpubkey/openpubkey/pull/221) I believe consent was needed for a refreshed ID Token.

Can you run an experiment to find out if consent is needed in Google/Azure/gitlab? If it is needed, please PR a comment explaining when and why it is needed, if it is not needed, I'll merge your PR as is.

kristiangronas:
Found a few issues about this https://github.com/googleapis/google-api-python-client/issues/213, it seems google has a weird quirk where you only get a refresh token if prompt=consent

From what i can tell google expects you to keep the refresh token that you got on the first login, but this doesn't seem easy

Since this is a google quirk maybe it makes sense to set prompt=consent just when the IP is google?

EthanHeilman:
> Since this is a google quirk maybe it makes sense to set prompt=consent just when the IP is google?

If we are sure that it doesn't cause problems for other OPs, then that makes sense to me.

kristiangronas:
GitLab does not work at all: https://gitlab.com/gitlab-org/gitlab/-/issues/377368, so it would also need to be special cased

Okta/Auth0 does not support account selection yet, but just ignores the parameter: https://community.auth0.com/t/let-user-choose-between-using-auth0-session-or-login-with-different-account/182094

Dex basically only does passthrough to another provider, but the parameter was ignored at least

Azure works well with select_account

Zitadel worked well too

EthanHeilman:
@kristiangronas That's a really good find that gitlab doesn't work if consent is set and the user is not logged into gitlab. We need to fix that

EthanHeilman:
Do you want to make all these changes in this PR?

EthanHeilman:
This PR allows you to specify the prompt via a config file. Do you want this feature in OpenPubkey or opkssh. If you want it in OpenPubkey I can port it here.

https://github.com/openpubkey/opkssh/pull/143

kristiangronas:
Yeah I mainly want it in opkssh, setting prompt=select_account is probably not necessary for most users, so i think it's an okay solution. We need to distribute a config file somehow anyway

EthanHeilman:
@kristiangronas Closing, please reopen if the client config in opkssh doesn't solve your issue.

PR #311, opened 2025-04-15
feat: discover signing algs to support es256

datosh:
Adds ability to use ES256 signatures for the OpenID Provider

Needed for: https://github.com/openpubkey/opkssh/issues/131


EthanHeilman (approved):
LGTM

PR #325, opened 2025-05-02
Support ClientSecret for AzureOpts

k0da:
Some tenants require to have client_secret to be specified

### DIFF
--- a/providers/azure.go
+++ b/providers/azure.go
@@ -32,6 +32,9 @@ type AzureOptions struct {
 	// ClientID is the client ID of the OIDC application. It should be the
 	// expected "aud" claim in received ID tokens from the OP.
 	ClientID string
+	// ClientSecret is the client secret of the OIDC application. Some OPs do
+	// not require that this value is set.
+	ClientSecret string
 	// Issuer is the OP's issuer URI for performing OIDC authorization and
 	// discovery.
 	Issuer string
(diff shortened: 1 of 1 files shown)


EthanHeilman:
Is this safe to do? Since the client secret in a OIDC public client makes the client secret public. What is the intention behind an azure tenants do this?

I'm open to merging this, but I want to make sure we allowing people to use unsafe configurations.

k0da:
@EthanHeilman 
Without supplying a client secret I'm getting:
```failed to exchange token: oauth2: "invalid_client" "AADSTS7000218: The request body must contain the following parameter: 'client_assertion' or 'client_secret'```

k0da:
@EthanHeilman I see. I have enabled Public flow it doesn't change error
```failed to exchange token: oauth2: "invalid_client" "AADSTS7000218: The request body must contain the following parameter: 'client_assertion' or 'client_secret'.```

I'm doing a POC around opkssh. And don't get it working yet. Yes. I'm building  a custom binary with additional debug messages to see whats going on there. Unfortunately I don't have full access to AzureAD, just created AppRegistration there.

EthanHeilman:
@k0da Thanks for looking into this. Are there any settings for AzureAD you can share so I can understand this better. I'm good with supporting the client secret in Azure, I just want to be sure it is safe to do so and actually solves the problem.

PR #329, opened 2025-05-09
Pass client secret opt to standard provider

kndndrj:
I think it's pretty self-explanatory.
I believe the client secret should be passed from opts to `StandardOp`.

I found that by using `github.com/oauth2-proxy/mockoidc` in one of my tests - this mock requires client_secret.

This fix makes it work (confirmed the behavior with keycloak as well).

### DIFF
--- a/providers/standard_provider.go
+++ b/providers/standard_provider.go
@@ -102,6 +102,7 @@ func GetDefaultStandardOpOptions(issuer string, clientID string) *StandardOpOpti
 func NewStandardOpWithOptions(opts *StandardOpOptions) BrowserOpenIdProvider {
 	return &StandardOp{
 		clientID:                  opts.ClientID,
+		ClientSecret:              opts.ClientSecret,
 		Scopes:                    opts.Scopes,
 		PromptType:                opts.PromptType,
 		AccessType:                opts.AccessType,


EthanHeilman:
Thanks for putting this PR together. I've been getting a lot of requests for this. 

- On what hand if you are using a public auth flow, which is the only auth flow that OpenPubkey supports the client secret should not be necessary. Allowing people to supply the client secret, may cause people to supply a client secret that should not be used in a public auth flow.

- On the other hand, OPs like google require the client secret, despite it being unnecessary and making it more likely that people will engage in insecure misconfigurations. Thankfully most OPs don't do this, but enough do that openpubkey has to care about it. This is a completely unforced error on the part of Google and some thing which I've seen cause issues for over a decade.

mockoidc, as you point out, does require a client secret despite not needing it for public flow and looks unwilling to fix this issue.  https://github.com/oauth2-proxy/mockoidc/issues/41

I've always felt that tools should take reasonable effects to protect people from misconfigurations, but not get in the way of user agency. Given how many OPs do this and often I am asked for this function, I think, despite my misgivings, it would be a mistake to not add the client-secret opt to the standard op.

EthanHeilman (approved):
LGTM

EthanHeilman:
Let me know if you want a release for this?

kndndrj:
Hey, thanks for the explanation. I fully agree that providing the secret is a bit strange and not wanted. But yeah, there are a few corner cases.

As for keycloak, I only tried it to make sure this change doesn't break anything.

Also, do you think it would make sense to state the fact that openpubkey is only supposed to work with public clients more obvious in the docs?

Thanks again and have a good day!

EthanHeilman:
> Also, do you think it would make sense to state the fact that openpubkey is only supposed to work with public clients more obvious in the docs?

Any OIDC client runs on the user's machine, like a single page javascript client or a desktop app has to be a public client. You can only have a confidential client in oidc if you are running the client on a protected server that the user has no control over.

Let me see what I can do to make the readme more explicit about this.

kndndrj:
You make a good point. I guess with that in mind adding this kind of info in the docs wouldn't really make much sense.

I have one more question - a bit off topic though, so I'm sorry about that, but...

Why wouldn't openpubkey support confidential clients? I know that 99.99% use cases are public clients, but wouldn't confidential clients have their own place too? For example something like an ssh proxy deployed on your server, that connects to remote devices in iot (just an example).

Or is the use of confidential clients just out of scope for openpubkey?

Anyways, thanks for all your answers! I appretiate your work on this project.

EthanHeilman:
> Why wouldn't openpubkey support confidential clients?

That is an excellent question and I don't have a good answer. The only answer I have is that no one has requested it yet and I personally don't need that usecase.

>  Or is the use of confidential clients just out of scope for openpubkey?

I'm open to someone adding that functionality. If someone openned an issue and said they really wanted to use openpubkey in production but they need confidential clients, I'd prioritize that feature. It shouldn't be that hard to add given a concrete usecase.

> For example something like an ssh proxy deployed on your server, that connects to remote devices in iot (just an example).

I can see that being useful. You have a webapp that you can briefly authorize the ability to run some SSH scripts by clicking a button and authorizing with Google.

EthanHeilman:
I needed a new release so your change is now in openpubkey 0.15.0

PR #336, opened 2025-06-22
feat: Add 12hr max age expiration policy

bmodotdev:
I believe this is the first step in supporting the downstream consumer feature of 12 hour expiration policy in opkssh i.e. https://github.com/openpubkey/opkssh/issues/229


bmodotdev:
Please let me know if you would also like the default expiration policy updated from 24hrs to 12hrs in [verifier.NewFromMany()](https://github.com/bmodotdev/openpubkey/blob/72512e6e2813ce67d7efae10de93f5449b0495c8/verifier/verifier.go#L120). I left the default at 24hr's to provide the least disruptive change.

EthanHeilman (approved):
LGTM, thanks for this PR

PR #348, opened 2025-11-22
Add configurable GQ audience prefix

kipz:
Makes the GQ audience prefix configurable instead of hard-coded to `"OPENPUBKEY-PKTOKEN:"`.

This is useful when the OIDC provider doesn't support setting an audience prefix in that format (e.g. Vault).

### DIFF
--- a/providers/providerverifier.go
+++ b/providers/providerverifier.go
@@ -36,10 +36,9 @@ import (
 const AudPrefixForGQCommitment = "OPENPUBKEY-PKTOKEN:"
 
 type DefaultProviderVerifier struct {
-	issuer      string
-	commitType  CommitType
-	options     ProviderVerifierOpts
-	gqAudPrefix string
+	issuer     string
+	commitType CommitType
+	options    ProviderVerifierOpts
 }
 
 type ProviderVerifierOpts struct {
@@ -78,12 +77,9 @@ func NewProviderVerifier(issuer string, options ProviderVerifierOpts) *DefaultPr
 		v.options.DiscoverPublicKey = discover.DefaultPubkeyFinder()
 	}
 
-	// Initialize GQ audience prefix
-	// Use the provided prefix from options, or default to the constant if empty
-	if options.GQAudiencePrefix == "" {
-		v.gqAudPrefix = AudPrefixForGQCommitment
-	} else {
-		v.gqAudPrefix = options.GQAudiencePrefix
+	// Initialize GQ audience prefix with default if not set
+	if v.options.GQAudiencePrefix == "" {
+		v.options.GQAudiencePrefix = AudPrefixForGQCommitment
 	}
 
 	return v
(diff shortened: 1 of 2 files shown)


EthanHeilman:
I was confused for a little bit because I didn't see the how the gitlab-ci aud value was being set in the provider, but the aud is configured in gitlab-ci so that doesn't matter

providers/providerverifier.go
```diff
@@ -73,6 +78,14 @@ func NewProviderVerifier(issuer string, options ProviderVerifierOpts) *DefaultPr
 		v.options.DiscoverPublicKey = discover.DefaultPubkeyFinder()
 	}
 
+	// Initialize GQ audience prefix
+	// Use the provided prefix from options, or default to the constant if empty
+	if options.GQAudiencePrefix == "" {
+		v.gqAudPrefix = AudPrefixForGQCommitment
+	} else {
+		v.gqAudPrefix = options.GQAudiencePrefix
+	}
```
EthanHeilman, line 83: Any reason to put this on the DefaultProviderVerifier struct? All the other options are kept in the options struct member.

Below we could then check if the `ProviderVerifierOpts.GQAudiencePrefix` is set and if not use the default.

Another benefit keeping it in options, is that we can check if it set in `VerifyIDToken`. If `GQAudiencePrefix` is set but this isn't a GQ ProviderVerifier then we fail and return an error.

If the check happens in `NewProviderVerifier` we can't return an error if we hit this misconfiguration because `NewProviderVerifier` doesn't return error.

providers/providerverifier_test.go
```diff
@@ -200,6 +200,107 @@ func TestProviderVerifier(t *testing.T) {
...
+			op, backendMock, _, err := NewMockProvider(providerOpts)
+			require.NoError(t, err)
+			opSignKey, keyID, _ := backendMock.RandomSigningKey()
+			idtTemplate.KeyID = keyID
+			idtTemplate.SigningKey = opSignKey
+
+			backendMock.SetIDTokenTemplate(&idtTemplate)
+
+			tokens, err := op.RequestTokens(context.Background(), cic)
+			require.NoError(t, err)
```
EthanHeilman:
```suggestion
			require.NoError(t, err, tc.name)
```

providers/providerverifier_test.go
```diff
@@ -200,6 +200,107 @@ func TestProviderVerifier(t *testing.T) {
...
+					CommitType:        CommitTypesEnum.GQ_BOUND,
+					ClientID:          clientID,
+					SkipClientIDCheck: true,
+					GQOnly:            true,
+					GQAudiencePrefix:  tc.gqAudiencePrefix,
+				},
+			}
+
+			op, backendMock, _, err := NewMockProvider(providerOpts)
+			require.NoError(t, err)
```
EthanHeilman:
```suggestion
			require.NoError(t, err, tc.name)
```

providers/providerverifier_test.go
```diff
@@ -200,6 +200,107 @@ func TestProviderVerifier(t *testing.T) {
...
+					GQAudiencePrefix:  tc.gqAudiencePrefix,
+				})
+
+			err = pv.VerifyIDToken(context.Background(), idToken, cic)
+
+			if tc.expError != "" {
+				require.ErrorContains(t, err, tc.expError)
+			} else {
+				require.NoError(t, err)
+			}
```
EthanHeilman, line 299:
```suggestion
			if tc.expError != "" {
				require.ErrorContains(t, err, tc.expError, tc.name)
			} else {
				require.NoError(t, err, tc.name)
			}
```

EthanHeilman (changes requested):
Looks pretty good, I have some minor comments

providers/providerverifier.go
kipz: You're right. That was silly - just didn't know my way around all this new code that well yet :) Should be fixed.

EthanHeilman (approved):
LGTM

PR #350, opened 2025-12-02
Add OP EdDSA signature algorithm support

kipz:
Adds EdDSA (Ed25519) signature algorithm support for OPs which is getting wider adoption these days (e.g. [keycloak](https://www.keycloak.org/2024/03/keycloak-2400-released)

### DIFF
--- a/client/choosers/mock_web_chooser.go
+++ b/client/choosers/mock_web_chooser.go
@@ -17,10 +17,14 @@
 package choosers
 
 import (
+	"errors"
 	"fmt"
+	"net"
 	"net/http"
+	"time"
 
 	"github.com/openpubkey/openpubkey/providers"
+	"github.com/sirupsen/logrus"
 )
 
 func NewMockWebChooser(opList []providers.BrowserOpenIdProvider, opToChoose string) *WebChooser {
@@ -35,7 +39,8 @@ func NewMockWebChooser(opList []providers.BrowserOpenIdProvider, opToChoose stri
 
 func BrowserOpenOverride(opToChoose string) func(string) error {
 	return func(uri string) error {
-		resp, err := http.Get(uri)
+		// Retry with exponential backoff to wait for server to be ready
+		resp, err := retryHTTPGet(uri)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


EthanHeilman:
@kipz It looked like a threading issue was causing the failures, I think it was unrelated to your change

discover/discover.go
```diff
@@ -42,34 +43,46 @@ type PublicKeyRecord struct {
...
-		pubKey = new(ecdsa.PublicKey)
-	} else if key.Algorithm().String() == "" {
+	// Let jwx handle the key extraction generically
+	var pubKey crypto.PublicKey
+	if err := key.Raw(&pubKey); err != nil {
+		return nil, fmt.Errorf("failed to decode public key: %w", err)
+	}
+
+	// If we got a private key (crypto.Signer), extract its public key
+	if signer, ok := pubKey.(crypto.Signer); ok {
```
EthanHeilman, line 54: I don't understand this step, shouldn't the prior ` key.Raw(&pubKey)` already get us a public key even if the type behind it is a crypto.Signer? Does doing this protect against leakage in case `key.Raw(&pubKey)` just wraps the signer?

What's the advantage of this approach?

EthanHeilman: Debugging this, it seems like it always creates a signer

providers/providerverifier.go
```diff
@@ -137,39 +138,39 @@ func (v *DefaultProviderVerifier) VerifyIDToken(ctx context.Context, idToken []b
...
 		pubKeyRecord, err := v.providerPublicKey(ctx, idToken)
 		if err != nil {
 			return fmt.Errorf("failed to get OP public key: %w", err)
 		}
 
-		// Ensure that the algorithm of public key from OpenID Provider matches the algorithm specified in the ID Token
-		if _, ok := pubKeyRecord.PublicKey.(*ecdsa.PublicKey); !ok {
-			return fmt.Errorf("public key is not an ECDSA public key")
+		// Validate that key type matches the algorithm in the token
+		switch alg {
```
EthanHeilman, line 153: This switch inside of a switch weirded me out at first, but I see what your doing here and it makes sense to me.

EthanHeilman (approved):

discover/discover.go
kipz: Yeah, so even if there's private key material, it successfully loads it into the public key! I've added a note for clarity, and I think we've covered it in the tests, and I've added a new one for RSA w/private key too.

EthanHeilman:
@kipz You know have release v0.20.0 https://github.com/openpubkey/openpubkey/releases/tag/v0.20.0

PR #351, opened 2025-12-02
Use CosignerVerifier interface instead of concrete type

kipz:
I'm sure this was the intention, but didn't come up until now:

was:

```golang
func WithCosignerVerifiers(verifiers ...*cosigner.DefaultCosignerVerifier) VerifierOpts
```

now:

```golang
func WithCosignerVerifiers(verifiers ...CosignerVerifier) VerifierOpts
```


EthanHeilman (approved):
LGTM

PR #361, opened 2026-03-28
feat: add GitHub Actions example, integration workflow, and harden CI

fdcastel:
## Summary

Implements the ideas discussed in #117 — adds a GitHub Actions example, a dedicated integration test workflow, and hardens existing CI pipelines.

### New: GitHub Actions Example (`examples/github_actions/`)
- **`example.go`**: End-to-end example using `providers.NewGithubOpFromEnvironment()` to generate and verify a PK Token via GitHub Actions OIDC. Follows the same pattern as the GitLab example.
- **`example_test.go`**: Unit test using a mock provider (runs with `go test ./...`).
- **`example_integration_test.go`**: Integration test behind `//go:build integration` tag. Only runs when explicitly invoked with `-tags integration` in a GitHub Actions environment with `id-token: write` permission.

### New: Integration Workflow (`.github/workflows/integration.yml`)
- Triggers on push to `main` (post-merge only — safe because only reviewed code lands on main).
- Scoped `id-token: write` permission for the integration job only.
- Runs `go test -tags integration -v ./examples/github_actions/`.

### Improved: CI Hardening
- **Job timeouts**: Added `timeout-minutes: 15` to test jobs in both `go.yml` and `staging.yml`.

## Note on `-race` detection

During development, I tested with `go test -race ./...` and found a **pre-existing data race** in `TestClientWithWebChooser` (`client/opkclient_test.go` — race between `StandardOp.defaultRequestTokens` and `StandardOp.ReuseBrowserWindowHook`). Enabling `-race` in CI would catch this and future races, but would block all PRs until the existing race is fixed. I've left it out for now to avoid disruption, but it may be worth addressing in a follow-up.

## Context

PR #117 identified the need for real GitHub Actions OIDC integration testing but was blocked by GitHub's security model (fork PRs cannot request ID tokens). The PR author concluded that an example + post-merge integration test was the best approach (Option 3). This PR implements that approach.

Closes #117

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -2,10 +2,6 @@ name: Go Checks
 
 on:
   push:
-    paths:
-      - "**.go"
-      - "go.mod"
-      - "go.sum"
   pull_request:
     paths:
       - "**.go"
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,6 +1,11 @@
 name: Go Checks
 
 on:
+  push:
+    paths:
+      - "**.go"
+      - "go.mod"
+      - "go.sum"
   pull_request:
     paths:
       - "**.go"
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -24,6 +24,7 @@ jobs:
   gotest:
     name: Run Tests
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,0 +1,28 @@
+name: Integration Tests
+
+on:
+  push:
+    branches: [ "main" ]
+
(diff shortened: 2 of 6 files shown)


.github/workflows/go.yml
```diff
@@ -1,6 +1,7 @@
 name: Go Checks
 
 on:
+  push:
```
EthanHeilman:
```suggestion
```

We already get all changes to PR by `pull_request:` I don't see any value to adding push here

.github/workflows/integration.yml
```diff
@@ -0,0 +1,28 @@
+name: Integration Tests
+
+on:
+  push:
+    branches: [ "main" ]
```
EthanHeilman:
```suggestion
    branches: [ "main" ] # SECURITY CRITICAL: DO NOT delete this line, DO NOT allow this to run on other branches. This integration tests can read github ID Tokens for this project.
```

EthanHeilman (approved):

PR #362, opened 2026-03-28
fix: remove duplicate JWKS fetch, propagate context, fix status code check

fdcastel:
## Summary

Fixes three bugs in `GetJwksByIssuer()` (`discover/discover.go`):

1. **Duplicate JWKS fetch** — The JWKS endpoint was fetched twice; the first response was discarded.
2. **Missing context propagation** — The second (actually-used) fetch used `httpClient.Get()` instead of `http.NewRequestWithContext()`, so the caller's `context.Context` (including timeouts and cancellation) did not apply.
3. **Wrong variable in status code check** — The error message referenced the response from the first (discarded) request instead of the second.
4. **Resource leak** — `resp.Body` from the second fetch was never closed.

The fix consolidates into a single context-aware fetch with proper `defer` cleanup.

Closes #360

## Note

This PR is based on #361 (the `fix-117` branch). It should be merged after #361 lands, or the base branch should be changed to `fix-117` if preferred.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -2,10 +2,6 @@ name: Go Checks
 
 on:
   push:
-    paths:
-      - "**.go"
-      - "go.mod"
-      - "go.sum"
   pull_request:
     paths:
       - "**.go"
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,6 +1,11 @@
 name: Go Checks
 
 on:
+  push:
+    paths:
+      - "**.go"
+      - "go.mod"
+      - "go.sum"
   pull_request:
     paths:
       - "**.go"
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -24,6 +24,7 @@ jobs:
   gotest:
     name: Run Tests
     runs-on: ubuntu-latest
+    timeout-minutes: 15
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
--- a/examples/github_actions/example.go
+++ b/examples/github_actions/example.go
@@ -26,19 +26,15 @@ import (
 	"github.com/openpubkey/openpubkey/verifier"
 )
 
-type Opts struct {
-	altOp providers.OpenIdProvider
+type GithubActionsExample struct {
+	// Op overrides the default GitHub Actions provider. Used for testing.
+	Op providers.OpenIdProvider
 }
 
-func SignWithGithubActions(opts ...Opts) error {
-	var op providers.OpenIdProvider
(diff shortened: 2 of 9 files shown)


.github/workflows/go.yml
```diff
@@ -24,6 +25,7 @@ jobs:
   gotest:
     name: Run Tests
     runs-on: ubuntu-latest
+    timeout-minutes: 15
```
EthanHeilman, line 28: Why these changes? Why test on push? Why 15 minutes? It seems like the tests take less than a minute?

examples/github_actions/example.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/openpubkey/openpubkey/client"
+	"github.com/openpubkey/openpubkey/providers"
+	"github.com/openpubkey/openpubkey/verifier"
+)
+
+type Opts struct {
+	altOp providers.OpenIdProvider
+}
+
+func SignWithGithubActions(opts ...Opts) error {
```
EthanHeilman: This function is called SignWithGithubActions, it doesn't need to support other OPs other than github

EthanHeilman: Ok, ok, you are doing this to mock this for the unittest. Do the mocking by setting the op as a member on the struct. ...opts is overkill here

EthanHeilman (changes requested):
Nice catch on this bug

.github/workflows/go.yml
fdcastel: These changes come from the parent PR #361 (`fix-117` branch), which this PR is stacked on top of. See [PR #361](https://github.com/openpubkey/openpubkey/pull/361) for the full rationale:

- **`push` trigger**: So that collaborators pushing to branches on their forks get CI feedback immediately, instead of only when opening a PR against upstream.
- **`timeout-minutes: 15`**: A safety net to prevent stuck jobs from consuming Actions minutes indefinitely. Tests normally complete in under a minute, but this guards against hangs (e.g., a test waiting on a network call that never returns). The 15-minute value is intentionally generous — it should never be hit under normal conditions.

I'm happy to adjust the timeout value if you'd prefer something tighter (e.g., 5 minutes).

examples/github_actions/example.go
fdcastel: Done. Replaced the `...Opts` variadic pattern with a `GithubActionsExample` struct that has an exported `Op` field. Tests set it directly for mocking; production code (integration test) leaves it `nil` to get the real GitHub Actions provider.

See commit 9778f8c.

EthanHeilman (approved):

PR #364, opened 2026-03-29
feat: cacheing mechanism for JWKS

ianroberts:
Introduce the concept of a *cache* for the JWKS discovery mechanism to avoid having to reach out and fetch the openid-configuration and JWKS URLs from the IdP on every single authentication attempt.  The verifier first attempts to fetch the issuer's JWKS from the cache and verify the token against that.  If there is no cache entry for that issuer, the cache entry is older than a (configurable) max age, or the cached key does not validate the signature, then it tries again with a freshly-fetched JWKS.  We include a simple in-memory implementation of the `DiscoveryCache` interface within openpubkey, primarily for use by tests.  It is expected that downstream consumers such as opkssh will include their own file- or database-based cache implementations providing longer term persistence and sharing of cached keys across processes.

Note that this changes the signatures of a number of public methods on `OpenIdProvider` and `discover.PublicKeyFinder`, to add a parameter for the caller to specify whether they want to allow use of the cache, and an extra return value saying whether the returned JWKS came from the cache or directly from the provider.  Even if the caller said that the key finder was *allowed* to use the cache, it may have gone out to the internet anyway if there was no suitable value in the cache, and we need to communicate this fact back to the caller.

This is the openpubkey side of the code to implement openpubkey/opkssh#440


discover/cache.go
```diff
@@ -0,0 +1,230 @@
...
+	return nil
+}
+
+// FilesystemDiscoveryCache is a DiscoveryCache implementation that uses a directory
+// on the filesystem for its cache store.  Entries are written to the cache by
+// creating files {BaseDir}/{SHA-256-of-issuer-URI}/jwks/jwks-{timestamp}-{random}
+// and read back by finding the corresponding file with the latest timestamp.
+// File modification times are not checked - the timestamp is the one encoded into
+// the file name.
+type FilesystemDiscoveryCache struct {
```
ianroberts: Having started looking at the `opkssh` side of this I'm now wondering if it makes more sense to put the `FilesystemDiscoveryCache` implementation in `opkssh` and leave just the interface and the dummy impls here, then the filesystem impl could use the same afero filesystem abstraction layer (which is a dependency of opkssh but not of openpubkey)

EthanHeilman: I agree, it makes a lot of sense to put the interface in opk and the implementation that opkssh needs in opkssh. The opkssh implementation is going to have to care about configuration and policy in a way that opk shouldn't.

We probably still want a simple implementation in opk for testing purposes and to make it easy for someone to use opk in other applications. Something like an in memory JWKS cache.

ianroberts: > We probably still want a simple implementation in opk for testing purposes and to make it easy for someone to use opk in other applications. Something like an in memory JWKS cache.

The `MapDiscoveryCache` implementation that I use in the tests is just a thin wrapper around a `map[string][]byte` but it ignores the maxAge parameter and doesn't do anything related to expiry.

ianroberts: I've removed the `FilesystemDiscoveryCache` implementation, it will go into opkssh instead.

EthanHeilman:
@ianroberts Create a PR on opkssh and I fix it up.

discover/discover.go
```diff
@@ -96,10 +97,22 @@ func DefaultPubkeyFinder() *PublicKeyFinder {
 	}
 }
 
+func NewPubkeyFinderWithCache(f JwksFetchFunc, cache DiscoveryCache, maxAge time.Duration) *PublicKeyFinder {
+	return &PublicKeyFinder{
+		JwksFunc:       f,
+		Cache:          cache,
+		StandardMaxAge: maxAge,
+		FallbackMaxAge: maxAge * 2,
```
EthanHeilman, line 105: FallbackMaxAge doesn't seem to be ever used in caching decisions and there is no test coverage of it.

discover/discover.go
```diff
@@ -133,49 +146,79 @@ func GetJwksByIssuer(ctx context.Context, issuer string, httpClient *http.Client
...
+	// if we get to here, either we were instructed not to use the cache, or
+	// there was no unexpired JWKS in the cache, or the JWKS that we pulled
+	// from the cache could not be parsed - go out to the provider
+	jwksJson, err = f.JwksFunc(ctx, issuer)
+	wasFallback := false
 	if err != nil {
-		return nil, fmt.Errorf(`failed to fetch JWKS: %w`, err)
+		// failed to fetch from provider, so see if the cache can provide a
+		// fallback possibly-expired key
+		var cacheErr any = struct{}{}
```
EthanHeilman:
```suggestion
		cacheErr := ErrCacheMiss // no cache configured or bad cache
```

cosigner/cosignerverifier.go
```diff
@@ -83,26 +83,38 @@ func (v *DefaultCosignerVerifier) VerifyCosigner(ctx context.Context, pkt *pktok
...
+func (v *DefaultCosignerVerifier) tryVerifyCosigner(ctx context.Context, pkt *pktoken.PKToken, header *pktoken.CosignerClaims, mayUseCache bool) (bool, error) {
+	keyRecord, wasCached, err := v.options.DiscoverPublicKey.ByKeyID(ctx, v.issuer, header.KeyID, mayUseCache)
+	if err != nil {
+		return wasCached, err
+	}
 	key := keyRecord.PublicKey
 	alg := keyRecord.Alg
 
 	// Check if it's expired
 	if time.Now().After(time.Unix(header.Expiration, 0)) {
```
EthanHeilman, line 106: We should check if the header is expired first before pulling from the jwks. No sense is looking up the JWKS is the cosigner header is expired. This PR doesn't introduce this bug, but we can fix it here.

cosigner/cosignerverifier.go
```diff
@@ -83,26 +83,38 @@ func (v *DefaultCosignerVerifier) VerifyCosigner(ctx context.Context, pkt *pktok
 		return fmt.Errorf("cosigner issuer (%s) doesn't match expected issuer (%s)", header.Issuer, v.issuer)
 	}
 
-	keyRecord, err := v.options.DiscoverPublicKey.ByKeyID(ctx, v.issuer, header.KeyID)
-	if err != nil {
+	wasCached, err := v.tryVerifyCosigner(ctx, pkt, header, true)
+	if !wasCached || err == nil {
```
EthanHeilman, line 87: Some errors like "cosigner signature expired" don't involve the cache and so we don't need to check the cache again. If ""cosigner signature expired" we should auto-fail, do not pass go, do not try without caching.

discover/discover.go
EthanHeilman, line 92: Add comment explaining that this used doesn't provide cache and looks up the public key fresh each time.

discover/discover.go
```diff
@@ -133,49 +146,79 @@ func GetJwksByIssuer(ctx context.Context, issuer string, httpClient *http.Client
...
 	}
 	jwks := jwk.NewSet()
 	if err := json.Unmarshal(jwksJson, jwks); err != nil {
-		return nil, fmt.Errorf(`failed to unmarshal JWKS: %w`, err)
+		return nil, false, fmt.Errorf(`failed to unmarshal JWKS: %w`, err)
 	}
-	return jwks, nil
+	if !wasFallback && f.Cache != nil {
+		_ = f.Cache.Write(issuer, jwksJson)
+		// TODO log if cache write fails
```
EthanHeilman: Add a OutOrErrWriter from utils and log this to errWriter


```
if err := f.Cache.Write(issuer, jwksJson); err!=nil {
  _, _ = fmt.Fprintf(errWriter(), "Failed to write to JWKS cache: %v\n", err)
}
```

providers/gq.go
```diff
@@ -75,7 +75,7 @@ func createGQTokenAllParams(ctx context.Context, idToken []byte, op OpenIdProvid
 		return nil, fmt.Errorf("gq signatures require ID Token signed with an RSA key, ID Token alg was (%s)", alg.String())
 	}
 
-	opKey, err := op.PublicKeyByToken(ctx, idToken)
+	opKey, _, err := op.PublicKeyByToken(ctx, idToken, true)
```
EthanHeilman, line 78: Shouldn't this retry as well, if it cached and fails, like the other PublicKey lookups?

providers/op.go
```diff
@@ -34,8 +34,8 @@ import (
 // an ID Token
 type OpenIdProvider interface {
 	RequestTokens(ctx context.Context, cic *clientinstance.Claims) (*simpleoidc.Tokens, error)
-	PublicKeyByKeyId(ctx context.Context, keyID string) (*discover.PublicKeyRecord, error)
-	PublicKeyByToken(ctx context.Context, token []byte) (*discover.PublicKeyRecord, error)
+	PublicKeyByKeyId(ctx context.Context, keyID string, mayUseCache bool) (*discover.PublicKeyRecord, bool, error)
```
EthanHeilman, line 37: Instead of passing mayUseCache as bool here, provide a function on the cache that lets you invalidate cache by issuer. Something like 

```
func (f *PublicKeyFinder) Invalidate(ctx context.Context, issuer string) error
```

this way we don't have to make a breaking chance to the OpenIdProvider interface and we can manage the cache via the PublicKeyFinder.

providers/op.go
```diff
@@ -34,8 +34,8 @@ import (
...
+	PublicKeyByToken(ctx context.Context, token []byte, mayUseCache bool) (*discover.PublicKeyRecord, bool, error)
```
EthanHeilman, line 38: Add the wasCachedBool to the `*discover.PublicKeyRecord` since it belongs in the PublicKeyRecord

discover/cache.go
```diff
@@ -0,0 +1,114 @@
...
+	// this would be time.Now but you may specify a different function for testing
+	// purposes via NewMapDiscoveryCacheWithClock
+	Now func() time.Time
+	// cache is the actual mapping from issuer to JWKS content
+	cache map[string][]byte
+	// timestamp records the time each entry was last written
+	timestamp map[string]time.Time
+	// mutex is used to avoid interference between read/write operations and calls
+	// to Expire
+	mutex *sync.Mutex
```
EthanHeilman, line 56:
```suggestion
	mutex sync.RWMutex
```

providers/standard_provider.go
```diff
@@ -103,6 +103,15 @@ type StandardOpOptions struct {
...
 	CallbackHTML string
+	// Cache is the discovery cache (if any) used by this provider
+	Cache discover.DiscoveryCache
+	// StandardMaxAge is the standard maximum age of a cache entry before it
+	// is considered expired
+	StandardMaxAge time.Duration
+	// FallbackMaxAge is the absolute maximum age of a cache entry - entries
+	// older than the StandardMaxAge but younger than the FallbackMaxAge may be
+	// used if the provider's JWKS endpoint cannot be reached
+	FallbackMaxAge time.Duration
```
EthanHeilman, line 114: Rather than add all of these to each provider we should just package it up. Something like

```
type CacheConfig struct {
    Cache          DiscoveryCache
    StandardMaxAge time.Duration
    FallbackMaxAge time.Duration
}
```

discover/cache.go
```diff
@@ -0,0 +1,114 @@
...
+// MapDiscoveryCache is a very simple in-memory cache implementation intended for
+// testing, or where a persistent cache is not required.  It maintains a map from
+// issuer ID to the retrieved JWKS, and the timestamp at which it was stored; a
+// call to Read will return the data that was written, unless the write timestamp
+// is older than the maxAge parameter in which case it will return a ErrCacheMiss error.
+type MapDiscoveryCache struct {
+	// Now is the function that is called to determine the current time.  Normally
+	// this would be time.Now but you may specify a different function for testing
+	// purposes via NewMapDiscoveryCacheWithClock
+	Now func() time.Time
```
EthanHeilman, line 49: Avoid exporting this so that someone doesn't mess with it and introduce hard to debug expiration errors
```suggestion
	now func() time.Time
```

discover/discover.go
```diff
@@ -133,49 +146,79 @@ func GetJwksByIssuer(ctx context.Context, issuer string, httpClient *http.Client
 	return io.ReadAll(response.Body)
 }
 
-func (f *PublicKeyFinder) fetchAndParseJwks(ctx context.Context, issuer string) (jwk.Set, error) {
-	jwksJson, err := f.JwksFunc(ctx, issuer)
+func (f *PublicKeyFinder) fetchAndParseJwks(ctx context.Context, issuer string, mayUseCache bool) (jwk.Set, bool, error) {
```
EthanHeilman: If the cache is corrupted and the JWKS is offline, this is will just return an error message about the JWKS failing to parse when the real error is that the JWKS is offline and can't replace the bad JWKS.

EthanHeilman (changes requested):
Looks pretty good, but has some blockers

discover/discover.go
EthanHeilman: Hope you don't mind, I fixed it

discover/discover.go
EthanHeilman: It is used now

ianroberts:
Thanks for the review, I will see how much of it I can address some time this weekend.

providers/op.go
ianroberts: A more fundamental question - why does `OpenIdProvider` have public methods for `PublicKeyBy...` in the first place?  As far as I can tell nothing in openpubkey or opkssh calls these methods _on the provider_, only ever on the `PublicKeyFinder` (and that only by `ProviderVerifier`); outside clients only use `RequestTokens` and `VerifyIDToken`, neither of which would need to expose anything related to cacheing.

EthanHeilman:
@ianroberts Whats the status of this?

providers/gq.go
ianroberts: This code path only runs as part of `opkssh login` on the _client_ side, and this PR is only concerned with cacheing for `opkssh verify` on the _server_ side.  There doesn't seem to be a lot of benefit in cacheing the JWKS on the client given the provider has to be reachable anyway in order to go through the auth flow.

providers/standard_provider.go
ianroberts: I've called it `DiscoveryCacheConfig` and put it in the `discover` package.

providers/op.go
ianroberts: I've implemented this, but there are a few situations where it's useful to know whether the cache was used or not even in error cases.  Therefore a number of places that used to return a `nil, err` now return an "empty" `PublicKeyRecord` (with just the `WasCached` value) instead of the `nil`.

providers/op.go
ianroberts: I've implemented this now, by making the `PublicKeyBy*` calls on a provider _always_ pass `mayUseCache = true` to the PKF.

cosigner/cosignerverifier.go
ianroberts: Good catch, fixed.

cosigner/cosignerverifier.go
ianroberts: This is implicitly resolved with the previous fix to check sig expiration first before fetching the pubkey - that particular error will now _always_ return `wasCached` as `false`, so won't retry.

ianroberts:
@EthanHeilman I think I've addressed all your previous comments, does this look a bit better now?

EthanHeilman:
@ianroberts Thanks let me take a look!

EthanHeilman (approved):
I made some small changes. I'm good to merge with this

PR #366, opened 2026-04-20
feat: Implements client-credentials flow

gautierrog:
Proposal for an implementation of client credentials flow.

Here is an example of a credential client requests with the response:

```
BASIC="$OPENPUBKEY_KEYCLOAK_CLIENT_ID:$OPENPUBKEY_KEYCLOAK_CLIENT_SECRET" curl -X POST "$OPENPUBKEY_KEYCLOAK_ISSUER/protocol/openid-connect/token" \
              -H "Content-Type: application/x-www-form-urlencoded" \
              -d "grant_type=client_credentials" \
              -d "scope=openid" \
              -d "client_id=$OPENPUBKEY_KEYCLOAK_CLIENT_ID" \
              -d "client_secret=$OPENPUBKEY_KEYCLOAK_CLIENT_SECRET" | jq
```

```json
{
  "access_token": "eyJhbGciOiJSUzI1NiIsInR5cCIgOiAiSldUIiwia2lkIiA6ICI1YWRlcUQ5WTZWUS1qMmFoZ3JvRktHSU1GdTRYaEJVd3lvdGpEejBKY2RJIn0.eyJleHAiOjE3ODAwNDIzNzMsImlhdCI6MTc4MDA0MjMxMywianRpIjoiZmNhZmM4MDktMjg1NC00ZGE4LWI4NjktZWY1MGJhNDIxYjdlIiwiaXNzIjoiaHR0cDovL2xvY2FsaG9zdDo4MDgwL3JlYWxtcy9tYXN0ZXIiLCJhdWQiOiJhY2NvdW50Iiwic3ViIjoiZjhlODlmMmMtODMyYS00NmIyLTg4NjctOGFiZjYxNTE1MGZiIiwidHlwIjoiQmVhcmVyIiwiYXpwIjoiT1BFTlBVQktFWS1QS1RPS0VOOiB0ZXN0LW0ybSIsImFjciI6IjEiLCJhbGxvd2VkLW9yaWdpbnMiOlsiLyoiXSwicmVhbG1fYWNjZXNzIjp7InJvbGVzIjpbImRlZmF1bHQtcm9sZXMtbWFzdGVyIiwib2ZmbGluZV9hY2Nlc3MiLCJ1bWFfYXV0aG9yaXphdGlvbiJdfSwicmVzb3VyY2VfYWNjZXNzIjp7ImFjY291bnQiOnsicm9sZXMiOlsibWFuYWdlLWFjY291bnQiLCJtYW5hZ2UtYWNjb3VudC1saW5rcyIsInZpZXctcHJvZmlsZSJdfX0sInNjb3BlIjoib3BlbmlkIGVtYWlsIHByb2ZpbGUiLCJlbWFpbF92ZXJpZmllZCI6ZmFsc2UsImNsaWVudEhvc3QiOiIxNzIuMjAuMC4xIiwicHJlZmVycmVkX3VzZXJuYW1lIjoic2VydmljZS1hY2NvdW50LW9wZW5wdWJrZXktcGt0b2tlbjogdGVzdC1tMm0iLCJjbGllbnRBZGRyZXNzIjoiMTcyLjIwLjAuMSIsImNsaWVudF9pZCI6Ik9QRU5QVUJLRVktUEtUT0tFTjogdGVzdC1tMm0ifQ.mxDD1AHXA4hogC52CacPAq2IS_SuR8lNQwDA2LrdmTQHxPtpiZrdLuVzdTCO9BzZh1Vw4pFCvU-l92bY1JujzE8rDQk21zIalXJsNegGwX1fCPJMVdQiXC-YDdvtsGey37ES95anPK3oXfPBykoaWfAgeEUzi9orGv1yaaJqEvmBjhvGZXTC_rCuntftTJ9dqkKyvjpGZ84cu28I5_7e1JrEgp-ev5gyZWOszi1nr_H-j_OG3piiQCVjJW9vfhSasFNmjX8Vijsm_VX9awzbPeTLD2cS0StFI0LUVfg0HIr7WG2lfbLyGD-uR9Ab63n-noVA3D2TdylLKoYSaYgsag",
  "expires_in": 60,
  "refresh_expires_in": 0,
  "token_type": "Bearer",
  "id_token": "eyJhbGciOiJSUzI1NiIsInR5cCIgOiAiSldUIiwia2lkIiA6ICI1YWRlcUQ5WTZWUS1qMmFoZ3JvRktHSU1GdTRYaEJVd3lvdGpEejBKY2RJIn0.eyJleHAiOjE3ODAwNDIzNzMsImlhdCI6MTc4MDA0MjMxMywiYXV0aF90aW1lIjowLCJqdGkiOiJkMzBkZWExNi0zMTE1LTRjNGEtOGY3ZS0yYWEzMzE2NmY1N2IiLCJpc3MiOiJodHRwOi8vbG9jYWxob3N0OjgwODAvcmVhbG1zL21hc3RlciIsImF1ZCI6Ik9QRU5QVUJLRVktUEtUT0tFTjogdGVzdC1tMm0iLCJzdWIiOiJmOGU4OWYyYy04MzJhLTQ2YjItODg2Ny04YWJmNjE1MTUwZmIiLCJ0eXAiOiJJRCIsImF6cCI6Ik9QRU5QVUJLRVktUEtUT0tFTjogdGVzdC1tMm0iLCJhdF9oYXNoIjoid0JzUlhpdmN0VGpGTmZGUVBXRlhHZyIsImFjciI6IjEiLCJlbWFpbF92ZXJpZmllZCI6ZmFsc2UsImNsaWVudEhvc3QiOiIxNzIuMjAuMC4xIiwicHJlZmVycmVkX3VzZXJuYW1lIjoic2VydmljZS1hY2NvdW50LW9wZW5wdWJrZXktcGt0b2tlbjogdGVzdC1tMm0iLCJjbGllbnRBZGRyZXNzIjoiMTcyLjIwLjAuMSIsImNsaWVudF9pZCI6Ik9QRU5QVUJLRVktUEtUT0tFTjogdGVzdC1tMm0ifQ.Op4IKlb9_1q8a4wrKXI_joBFsduUPGtc_3qsEp5RMrtGRwZDt69LqE9TcMX41vGY7hOrzfmXws_zm4N-MVYJdC0x_gRFUK3dLxgogjyenOQonBeIQ9rS89nxuKdNVF2vetMfkFqetP9oWUbRjeME-Tez3cSUCpw3lxvFmjf4kXAL9Xc24TqjRuzdPgj8GyGLgfI_0gmnnh_Gqq_uqQqBTrGEujDWMoFlqUN9oEzMeUxPiPu-QeY4vPTDRbOiIQxXjISNMgmX14-J9FA7lHUZtfeU1eL7_mTNmHoiwTIHqQjkH3hWndKDFWgXR2r0YYeBctmeBCYqlzB4jRf_VGZQyg",
  "not-before-policy": 0,
  "scope": "openid email profile"
}
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -222,18 +222,26 @@ func (o *OpkClient) oidcAuth(
 	if err != nil {
 		return nil, fmt.Errorf("error requesting OIDC tokens from OpenID Provider: %w", err)
 	}
-	idToken := tokens.IDToken
+	providerToken := tokens.IDToken
+	if len(providerToken) == 0 {
+		if _, ok := o.Op.(providers.ClientCredentialsOpenIdProvider); ok && len(tokens.AccessToken) > 0 {
+			providerToken = tokens.AccessToken
+		}
+	}
+	if len(providerToken) == 0 {
+		return nil, fmt.Errorf("provider response missing ID token")
+	}
 	o.refreshToken = tokens.RefreshToken
 	o.accessToken = tokens.AccessToken
 
 	// Sign over the payload from the ID token and client instance claims
-	cicToken, err := cic.Sign(signer, alg, idToken)
+	cicToken, err := cic.Sign(signer, alg, providerToken)
 	if err != nil {
 		return nil, fmt.Errorf("error creating cic token: %w", err)
 	}
 
 	// Combine our ID token and signature over the cic to create our PK Token
-	pkt, err := pktoken.New(idToken, cicToken)
+	pkt, err := pktoken.New(providerToken, cicToken)
 	if err != nil {
 		return nil, fmt.Errorf("error creating PK Token: %w", err)
 	}
--- a/client/client.go
+++ b/client/client.go
@@ -222,26 +222,18 @@ func (o *OpkClient) oidcAuth(
 	if err != nil {
 		return nil, fmt.Errorf("error requesting OIDC tokens from OpenID Provider: %w", err)
 	}
-	providerToken := tokens.IDToken
-	if len(providerToken) == 0 {
-		if _, ok := o.Op.(providers.ClientCredentialsOpenIdProvider); ok && len(tokens.AccessToken) > 0 {
-			providerToken = tokens.AccessToken
-		}
-	}
-	if len(providerToken) == 0 {
-		return nil, fmt.Errorf("provider response missing ID token")
-	}
+	idToken := tokens.IDToken
 	o.refreshToken = tokens.RefreshToken
 	o.accessToken = tokens.AccessToken
 
 	// Sign over the payload from the ID token and client instance claims
-	cicToken, err := cic.Sign(signer, alg, providerToken)
+	cicToken, err := cic.Sign(signer, alg, idToken)
 	if err != nil {
 		return nil, fmt.Errorf("error creating cic token: %w", err)
 	}
 
 	// Combine our ID token and signature over the cic to create our PK Token
-	pkt, err := pktoken.New(providerToken, cicToken)
+	pkt, err := pktoken.New(idToken, cicToken)
 	if err != nil {
 		return nil, fmt.Errorf("error creating PK Token: %w", err)
 	}
--- a/providers/standard_provider.go
+++ b/providers/standard_provider.go
@@ -19,7 +19,9 @@ package providers
 import (
 	"context"
 	"crypto"
+	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"strings"
@@ -33,6 +35,7 @@ import (
 	"github.com/openpubkey/openpubkey/providers/qrcode"
 	"github.com/openpubkey/openpubkey/util"
 	"github.com/sirupsen/logrus"
+	oidcclient "github.com/zitadel/oidc/v3/pkg/client"
 	"github.com/zitadel/oidc/v3/pkg/client/rp"
 	oidchttp "github.com/zitadel/oidc/v3/pkg/http"
 	"github.com/zitadel/oidc/v3/pkg/oidc"
@@ -84,6 +87,9 @@ type StandardOpOptions struct {
 	GQSign bool
 	// DeviceFlow denotes if the client should use the device flow instead of the standard
 	DeviceFlow bool
+	// ClientCredentialsFlow denotes if the client should use the client credentials
+	// flow to request tokens when RequestTokens is called.
+	ClientCredentialsFlow bool
 	// OpenBrowser denotes if the client's default browser should be opened
 	// automatically when performing the OIDC authorization flow. This value
 	// should typically be set to true, unless performing some headless
@@ -115,13 +121,14 @@ func GetDefaultStandardOpOptions(issuer string, clientID string) *StandardOpOpti
 			"http://localhost:10001/login-callback",
 			"http://localhost:11110/login-callback",
 		},
-		RemoteRedirectURI: "",
-		GQSign:            false,
-		DeviceFlow:        false,
-		OpenBrowser:       true,
-		HttpClient:        nil,
-		IssuedAtOffset:    1 * time.Minute,
-		CallbackHTML:      defaultCallbackHTML,
+		RemoteRedirectURI:     "",
+		GQSign:                false,
+		DeviceFlow:            false,
+		ClientCredentialsFlow: false,
+		OpenBrowser:           true,
+		HttpClient:            nil,
+		IssuedAtOffset:        1 * time.Minute,
+		CallbackHTML:          defaultCallbackHTML,
 	}
 }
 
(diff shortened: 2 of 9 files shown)


client/client.go
```diff
@@ -222,18 +222,26 @@ func (o *OpkClient) oidcAuth(
 	if err != nil {
 		return nil, fmt.Errorf("error requesting OIDC tokens from OpenID Provider: %w", err)
 	}
-	idToken := tokens.IDToken
+	providerToken := tokens.IDToken
```
EthanHeilman: Why set the ID Token with the Access Token? We shouldn't be treating ID Tokens as if they are access tokens.

providers/standard_provider.go
```diff
@@ -390,18 +408,108 @@ func (s *StandardOp) RequestTokens(ctx context.Context, cic *clientinstance.Clai
...
+	var tokenResponse struct {
+		AccessToken  string `json:"access_token"`
+		RefreshToken string `json:"refresh_token"`
+		IDToken      string `json:"id_token"`
+	}
+	if err := json.Unmarshal(body, &tokenResponse); err != nil {
+		return nil, fmt.Errorf("failed to decode token response: %w", err)
+	}
+
+	if tokenResponse.AccessToken == "" && tokenResponse.IDToken == "" {
```
EthanHeilman: Is an access token always required for client credentials?

providers/standard_provider.go
```diff
@@ -390,18 +408,108 @@ func (s *StandardOp) RequestTokens(ctx context.Context, cic *clientinstance.Clai
...
+	}
+
+	return &simpleoidc.Tokens{
+		IDToken:      nilIfEmpty(tokenResponse.IDToken),
+		RefreshToken: nilIfEmpty(tokenResponse.RefreshToken),
+		AccessToken:  nilIfEmpty(tokenResponse.AccessToken),
+	}, nil
+}
+
+func (s *StandardOp) clientCredentialsRequestTokens(ctx context.Context, _ string) (*simpleoidc.Tokens, error) {
```
EthanHeilman: Uppercase struct function names are exported, lowercase are unexported.  It doesn't seem like you need RequestClientCredentialsTokens to be exported. You can probably just have clientCredentialsRequestTokens as the function and drop RequestClientCredentialsTokens.

providers/standard_provider.go
```diff
@@ -553,11 +661,24 @@ func (s *StandardOp) ClientID() string {
...
 func (s *StandardOp) VerifyIDToken(ctx context.Context, idt []byte, cic *clientinstance.Claims) error {
+	commitType := CommitTypesEnum.NONCE_CLAIM
+	gqOnly := false
+	skipClientIDCheck := false
+
+	if s.ClientCredentialsFlow {
+		commitType = CommitTypesEnum.GQ_BOUND
+		gqOnly = true
+		// GQ commitment verification requires skipping strict client-id/audience check.
+		skipClientIDCheck = true
```
EthanHeilman, line 667: Does GQ commitment verification requires skipping strict client-id/audience check in this case? Does the ID Token not have an audience scoping?

EthanHeilman (changes requested):
If you write unittests for this flow and address the comments, I think I can merge this

client/client.go
gautierrog: the idea was that client credentials doesn't seem to guarantee providing any id token. after further checks it seems to be implementation dependant.
I'll take a step back on this behavior and rely on ID token as it is initially done.

providers/standard_provider.go
gautierrog: afaik it should always be present, although it's not mandatory for openpubkey, I'll loosen the check

providers/standard_provider.go
gautierrog: This is to follow the sanity checks on https://github.com/openpubkey/openpubkey/blob/main/providers/providerverifier.go#L108

The check says that "[...] the audience can't be the client-id" which I honnestly don't get quite well. Maybe the sanity check about skipping client id check while in GQ commitment is not really needed. What do you think @EthanHeilman ?

providers/standard_provider.go
gautierrog: OK. I removed the exported function

gautierrog:
I wrote unit tests in 

https://github.com/openpubkey/openpubkey/pull/366/changes#diff-c815ce22e5e4f1ec054611a560486e5448c31d3cd3ebe7c641e562def70d874dR378
and

 https://github.com/openpubkey/openpubkey/pull/366/changes#diff-43e8423bcc9bb4d9fcd3ce8da7e0197e64d49376e957292eadba0316d57b000dR343
 
 did you have other places to write tests in mind?

providers/standard_provider.go
EthanHeilman: The comment I wrote is confusingly written:

```
// When we bind the commitment to the ID Token using GQ Signatures,
// We require that the audience is prefixed with
// "OPENPUBKEY-PKTOKEN:"
```
Github and gitlab:
1. in github and gitlab the client can set any aud it wants,
2. in github and gitlab there is no notion of a client-id
3. github and gitlab require GQ signatures to do the pubkey binding

Because github and gitlab are using a non-standardized flow that they copied from firebase

This differs from typical OpenID Connect where
1. the aud MUST be the client-id

Because github and gitlab are the only providers that use GQ signatures and because they don't have client-id I am conflating the idea of no client-id with GQ signatures. In your case the ID Token may in fact have a client-id.

Could you paste an example client credential ID Token into the PR description? It would help me understand example what you are trying to do

gautierrog: I updated the PR description in accordance

PR #369, opened 2026-06-15
Add PS256 (RSASSA-PSS) algorithm support in publicKeyRecordFromJWK

kipz:
## Summary

- `publicKeyRecordFromJWK` in `discover/discover.go` now accepts PS256 (RSASSA-PSS) in addition to RS256 when the JWK key type is RSA

Both algorithms use RSA public keys, so the validation logic is identical — they share a single `case` arm.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


providers/mocks/idtoken.go
```diff
@@ -128,7 +128,11 @@ func (t *IDTokenTemplate) IssueTokensWithSubject(subject *Subject) (*oidc.Tokens
 
 	var providerAlg jwa.KeyAlgorithm
 	if _, ok := t.SigningKey.Public().(*rsa.PublicKey); ok {
-		providerAlg = jwa.RS256()
+		if t.Alg == "PS256" {
+			providerAlg = jwa.PS256()
+		} else {
+			providerAlg = jwa.RS256()
+		}
```
EthanHeilman, line 137: can you throw a sanity check on here to catch typos. Like 

```
if t.Alg == "PS256" {
			providerAlg = jwa.PS256()
		} if t.Alg == "RS256" {
            providerAlg = jwa.RS256()
else {
			return error("bad thing has happened 0_0")
		}
```

EthanHeilman:
This looks pretty good, is there anything weird that happens with PS256 and GQ? Do we need to make any changes there? Does PS256 work out of the box?

kipz:
> This looks pretty good, is there anything weird that happens with PS256 and GQ? Do we need to make any changes there? Does PS256 work out of the box?

@EthanHeilman good question. I'll make sure this is tested.

providers/mocks/idtoken.go
kipz: Done - added an explicit `RS256` branch and returns an error for any unexpected RSA algorithm.

EthanHeilman (approved):

EthanHeilman:
Thanks for this!

kipz:
@EthanHeilman - feel a bit stupid now, but PSS uses a salt (which is recoverable from the signed message), but of course we throw it away when using GQ. We can fix it by hanging on to the salt in protected headers, but it might make sense to throw an error until we have that figured out. WDYT?

EDIT: I see we do throw an error :)

EthanHeilman:
I think throw an error until we have a good plan on the best approach here

PR #371, opened 2026-06-15
fix: support RSA exponents other than 65537 for GQ signatures

### LINKED ISSUE #230: GQ library should support RSA exponents other than 2^16+1 safely

kipz:
Fixes #230.

The `t` parameter was computed incorrectly for exponents other than 65537, giving far fewer rounds than the security parameter requires. This uses `floor(log2(E))` bits per round and removes the 65537-only guard.

### DIFF
--- a/gq/gq.go
+++ b/gq/gq.go
@@ -135,6 +135,9 @@ func NewSignerVerifier(publicKey *rsa.PublicKey, securityParameter int) (SignerV
 	if publicKey.E < 3 || publicKey.E%2 == 0 {
 		return nil, fmt.Errorf("unsupported RSA public key exponent: %d, must be odd and at least 3", publicKey.E)
 	}
+	if securityParameter <= 0 {
+		return nil, fmt.Errorf("security parameter must be positive, got %d", securityParameter)
+	}
 	n, v, nBytes, vBytes, err := parsePublicKey(publicKey)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse RSA public key: %w", err)
--- a/gq/gq.go
+++ b/gq/gq.go
@@ -132,8 +132,8 @@ func New256SignerVerifier(publicKey *rsa.PublicKey) (SignerVerifier, error) {
 //
 // The securityParameter parameter is the level of desired security in bits. 256 is recommended.
 func NewSignerVerifier(publicKey *rsa.PublicKey, securityParameter int) (SignerVerifier, error) {
-	if publicKey.E < 3 {
-		return nil, fmt.Errorf("unsupported RSA public key exponent: %d, must be at least 3", publicKey.E)
+	if publicKey.E < 3 || publicKey.E%2 == 0 {
+		return nil, fmt.Errorf("unsupported RSA public key exponent: %d, must be odd and at least 3", publicKey.E)
 	}
 	n, v, nBytes, vBytes, err := parsePublicKey(publicKey)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


gq/gq.go
```diff
@@ -131,15 +132,19 @@ func New256SignerVerifier(publicKey *rsa.PublicKey) (SignerVerifier, error) {
...
 	}
 	n, v, nBytes, vBytes, err := parsePublicKey(publicKey)
-	t := securityParameter / (vBytes * 8)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse RSA public key: %w", err)
+	}
+	// Each GQ1 round provides floor(log2(E)) bits of soundness. Compute t = ceil(securityParameter / floor(log2(E))).
+	// bits.Len(uint(E)) - 1 equals floor(log2(E)) exactly for all E >= 2, avoiding floating-point imprecision.
+	bitsPerRound := bits.Len(uint(publicKey.E)) - 1
+	t := (securityParameter + bitsPerRound - 1) / bitsPerRound
```
EthanHeilman, line 148: I don't understand this equation, should it be ` t := (securityParameter) / bitsPerRound`  Why is bitsPerRound beign added?

kipz: it computes the ceiling, rather than the floor, which could give fewer rounds

EthanHeilman: ahh! That makes sense

EthanHeilman (approved):
LGTM

Actually can you add a test that does sign and verify but with e=3 just as a sanity check

kipz:
> Actually can you add a test that does sign and verify but with e=3 just as a sanity check

@EthanHeilman - threw llm at that. RSA key gen looks OK, but definitely worth a more detailed look. Maybe we can find a library to do that for us?

EthanHeilman (approved):

EthanHeilman (approved):
LGTM

EthanHeilman:
@kipz Added a comment about the function for generating test RSA keys so that we don't get security reports about it or so some LLM doesn't supply it as an answer in setting where security matters.

EthanHeilman:
@kips for this PR, I've been meaning to get around to addressing this for 3 years. Feels good to have to it closed out!

PR #372, opened 2026-06-19
Support GQ signatures over PS256 (RSASSA-PSS) JWTs

kipz:
## Summary

Extends PS256 (RSASSA-PSS) support to the GQ signing/verification path, which previously rejected PS256.

GQ strips the OP's RSA signature `s`, but PSS encoding mixes a random salt that lives inside `s`. The salt is still available at GQ-signing time, so it is recovered (`EM = s^e mod n`, then EMSA-PSS decode) and carried in the GQ protected header as `pss_salt`. The verifier reconstructs the encoded message `G` via EMSA-PSS encode from the signing payload and the carried salt.

`G` is always recomputed from the message, so tampering with the payload, the preserved original headers, or the carried salt changes `G` and fails verification. The original signature `s` remains stripped, preserving GQ's anti-replay property. The carried salt reveals nothing about `s` or the OP private key.

> Note: the EMSA-PSS code in `gq/pss.go` is hand-rolled (no stdlib export exists) and is security-sensitive — it warrants explicit maintainer/security review. It is pinned to the stdlib's `SignPSS`/`VerifyPSS` behaviour by `gq/pss_test.go`.


EthanHeilman:
@kipz Whats the status of this?

PR #373, opened 2026-06-24
fix: avoid panic on array aud in GQCommitment verification

kanywst:
`verifyCommitment` reads `aud` from a raw `map[string]any` and asserts `aud.(string)` (`providers/providerverifier.go:230`). RFC 7519 allows `aud` to be a string array, so a GQ-signed PK token whose ID token carries an array `aud` panics the verifier.

`oidc.OidcClaims.UnmarshalJSON` already handles array `aud`, so only the GQCommitment path is affected. This asserts the type and returns an error instead. `TestGQCommitmentArrayAudReturnsError` panics before the change and passes after.


kipz (approved):
LGTM

EthanHeilman (approved):

PR #374, opened 2026-07-03
Fixes browser launching on WSL

sukujgrg:
Fixes browser launching on WSL

## Problem

- **WSL UNC path warning**: `cmd /c start`, when run from a WSL process, inherits a UNC working directory that `cmd.exe` can't handle, printing a warning on every launch even though the browser still opens: 

``` 
CMD.EXE was started with the above path as the current directory. 
UNC paths are not supported.  Defaulting to Windows directory. 
```

- **`xdg-utils` dependency on WSL**: WSL images often don't have a desktop environment, so `xdg-utils` (`xdg-open`) isn't reliably available as a fallback for launching a browser.

## Changes

- Windows and WSL now launch via `powershell.exe -NoProfile -Command Start-Process <url>` instead of `cmd /c start` — avoids the UNC path warning and the `xdg-utils` dependency.
- Added `isWSL()` (checks `/proc/sys/kernel/osrelease` for `microsoft`/`wsl`) to route WSL to the PowerShell path.
- macOS (`open`) and native Linux (`xdg-open`) unchanged.

## Testing

- Verified on WSL (Ubuntu 24.04): UNC warning reproduced with `cmd.exe`, gone with PowerShell.
- Verified browser still opens correctly on macOS and native Linux.

### DIFF
--- a/util/launch.go
+++ b/util/launch.go
@@ -17,15 +17,18 @@
 package util
 
 import (
+	"encoding/base64"
+	"encoding/binary"
+	"fmt"
 	"os"
 	"os/exec"
 	"runtime"
 	"strings"
+	"unicode/utf16"
 )
 
 // https://stackoverflow.com/questions/39320371/how-start-web-server-to-open-page-in-browser-in-golang
 // open opens the specified URL in the default browser of the user.
-// url must be trusted: on Windows/WSL it is passed to powershell.exe -Command, which is vulnerable to injection if url contains untrusted input.
 func OpenUrl(url string) error {
 	switch runtime.GOOS {
 	case "windows":
@@ -42,14 +45,24 @@ func OpenUrl(url string) error {
 	}
 }
 
+// openWithPowerShell launches url via PowerShell using -EncodedCommand.
+// The URL is embedded as a PowerShell single-quoted string literal with
+// embedded single quotes escaped. This avoids injection through PowerShell
+// metacharacters in the URL. -EncodedCommand is used instead of -Command
+// because -Command re-joins and re-parses all trailing argv entries as a
+// single script, which would otherwise require additional care to avoid
+// argv-splitting ambiguity.
 func openWithPowerShell(url string) error {
-	return exec.Command(
-		"powershell.exe",
-		"-NoProfile",
-		"-Command",
-		"Start-Process",
-		url,
-	).Start()
+	script := fmt.Sprintf("Start-Process '%s'", strings.ReplaceAll(url, "'", "''"))
+
+	u16 := utf16.Encode([]rune(script))
+	buf := make([]byte, len(u16)*2)
+	for i, r := range u16 {
+		binary.LittleEndian.PutUint16(buf[i*2:], r)
+	}
+	encoded := base64.StdEncoding.EncodeToString(buf)
+
+	return exec.Command("powershell.exe", "-NoProfile", "-EncodedCommand", encoded).Start()
 }
 
 func isWSL() bool {
--- a/util/launch.go
+++ b/util/launch.go
@@ -20,6 +20,7 @@ import (
 	"encoding/base64"
 	"encoding/binary"
 	"fmt"
+	"net/url"
 	"os"
 	"os/exec"
 	"runtime"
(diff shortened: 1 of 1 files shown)


EthanHeilman:
@sukujgrg This looks pretty good, only concern is injection into powershell. Can you add code that prevents command-injection? It isn't security critical currently because the URI value is trusted and should not be attacker controlled, but I'd rather build security components so I don't have to worry about how they are composed.

EthanHeilman:
@sukujgrg I see you made some changes, ping me on here when you want to me rereview and merge

sukujgrg:
Done.
Tested func independently too:

image
image


```
package main

import (
        "fmt"
        "os"

        "github.com/openpubkey/openpubkey/util"
)

func main() {
        urls := os.Args[1:]
        if len(urls) == 0 {
                urls = []string{"https://example.com"}
        }

        for _, u := range urls {
                fmt.Printf("OpenUrl(%q) ... ", u)
                if err := util.OpenUrl(u); err != nil {
                        fmt.Printf("ERROR: %v\n", err)
                } else {
                        fmt.Println("ok")
                }
        }
}
```

util/launch.go
```diff
@@ -17,25 +17,60 @@
...
+}
+
+// openWithPowerShell launches url via PowerShell using -EncodedCommand.
+// The URL is embedded as a PowerShell single-quoted string literal with
+// embedded single quotes escaped. This avoids injection through PowerShell
+// metacharacters in the URL. -EncodedCommand is used instead of -Command
+// because -Command re-joins and re-parses all trailing argv entries as a
+// single script, which would otherwise require additional care to avoid
+// argv-splitting ambiguity.
+func openWithPowerShell(url string) error {
```
EthanHeilman, line 76: verify that the URI is actually a URI, i.e. starts with "http:" or "https:" to avoid someone passing `url=C:\Windows\System32\calc.exe`

util/launch.go
```diff
@@ -17,25 +17,60 @@
...
+	script := fmt.Sprintf("Start-Process '%s'", strings.ReplaceAll(url, "'", "''"))
```
EthanHeilman: Unfortunately microsoft did was microsoft always does and over complicated single quotes, you there are a bunch of bypasses to single quote escaping. https://clymb3r.wordpress.com/2013/09/07/avoiding-powershell-command-injection-unicode-issues/

I believe there is a pretty reliable fix all of this, which is to set the URI as an env variable. If you do that, I don't think it gets parsed.

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_parsing?view=powershell-7.6

util/launch.go
sukujgrg: This one did you mean to make it applicable for all platforms? Will it cause any breakage to someone if configured it with example.com/login

sukujgrg: It seems safe to do so but just confirming scope (whether for all platforms or just Windows) before pushing change.

```
[wslBrowserOpen *=] % grep -E 'OpenUrl|loginURI' providers/standard_provider.go client/choosers/web_chooser.go | grep -v logrus
providers/standard_provider.go: loginURI := fmt.Sprintf("http://localhost:%s/login", redirectURI.Port())
providers/standard_provider.go:         s.reuseBrowserWindowHook <- loginURI
providers/standard_provider.go:         if err := util.OpenUrl(loginURI); err != nil {
providers/standard_provider.go:         if err := s.browserOpenOverride(loginURI); err != nil {
client/choosers/web_chooser.go:         var loginURI string
client/choosers/web_chooser.go:                 loginURI = fmt.Sprintf("http://localhost:%d/chooser", port)
client/choosers/web_chooser.go:                 loginURI = fmt.Sprintf("http://%s/chooser", listener.Addr().String())
client/choosers/web_chooser.go:                 if err := util.OpenUrl(loginURI); err != nil {
client/choosers/web_chooser.go:                 if err := wc.browserOpenOverride(loginURI); err != nil {
```

EthanHeilman: Not sure what you mean? In all cases shouldn't it always be "https://..." or "http://..."?

> This one did you mean to make it applicable for all platforms?

I'd just do it for WSL, since WSL is the only one that allows command injection. All the other ones explicitly open a web browser.

util/launch.go
sukujgrg: Updated to use env var.

util/launch.go
```diff
@@ -17,25 +17,93 @@
...
+	if err := validateHTTPURL(url); err != nil {
+		return err
+	}
+
+	// Internal-only pass-through, not a supported/documented configuration
+	// variable: it exists solely to hand url to the child PowerShell process
+	// without it ever being parsed as script text. Set on cmd.Env below, so
+	// it is scoped to that one subprocess and never touches this process's
+	// or any user's actual environment.
+	const envVar = "OpenPubkeyBrowserURL"
```
sukujgrg, line 86: Used PowerShell style environment variable casing as it is defined [here](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/start-process?view=powershell-5.1#example-7-specifying-arguments-to-the-process).

EthanHeilman (approved):
LGTM - I tested this branch locally on regular windows (not WSL) using the example/web code

EthanHeilman:
Thanks for your work on this? Do you need a release?

sukujgrg:
Thank you. No problem. I'm happy to wait and have this change included in your next release of updates.

PR #375, opened 2026-07-07
providers: bind OIDC callback to the redirect URI host, not hardcoded localhost

sukujgrg:
FindAvailablePort (`func` in `providers/utils.go`) always bound "localhost:<port>", regardless of the configured redirect URI, and the login URL was opened on "localhost" too. "localhost" can resolve to both 127.0.0.1 and ::1, so the listener binds a single address family while the browser (which often prefers ::1) may connect to the other, silently dropping the callback.

Bind (and open /login on) the redirect URI's actual host so callers can pin a loopback IP literal (e.g. 127.0.0.1) for a deterministic single-family bind the browser is guaranteed to match. This also aligns with RFC [8252](https://datatracker.ietf.org/doc/html/rfc8252#section-8.3) (OAuth 2.0 for Native Apps) 8.3, which recommends loopback IP literals over "localhost" precisely to avoid inadvertently listening on interfaces other than loopback and to avoid issues from misconfigured hostname resolution.

Harden host validation while here: replace the HasPrefix checks with an exact "localhost" match plus net.ParseIP().IsLoopback(). The prefix check accepted lookalikes such as "localhost.attacker.com" and "127.0.0.1.attacker.com", which resolve via DNS to attacker-controlled addresses; since the host feeds both the listener bind and the browser navigation, that could steer either off loopback. Also check Hostname() rather than Host so bracketed IPv6 literals ([::1]) validate correctly.

Add tests covering the IPv4/IPv6 literal bind and rejection of non-loopback hosts, and document that redirect URIs must include an explicit port.

The problem that is trying to solve here can be simulated by doing this:

In one terminal open ipv6 port on all interfaces:

```
% nc -6 -k -l 49152
```

On second terminal run

```
nc -k -l 49152
```

On third terminal, start login flow (assume redirect url is `http://localhost:49152`)

Here browser will spin for ever because browser is resolving localhost to `::1` and on `::1` nc is listneing

lsof confirm:

```
% lsof -n -i:49152
COMMAND     PID   USER    FD   TYPE            DEVICE SIZE/OFF NODE NAME
nc        50906   suku    3u  IPv6 0xa6e11a5c23b99ad7      0t0  TCP *:49152 (LISTEN)
nc        53743   suku    3u  IPv4 0x45dbf57bed4927bd      0t0  TCP *:49152 (LISTEN)
opkdd     53790   suku    4u  IPv4 0x58d4f6f3b30e7aa3      0t0  TCP 127.0.0.1:49152 (LISTEN)
```

(Issue was noticed with Apple's rapportd deamon)

**Scope**

This change does not remove the address-family ambiguity when a redirect URI still uses `localhost`. It enables callers to configure an explicit loopback IP, such as `127.0.0.1` or `[::1]`, and ensures that both the listener and browser navigation use that exact address. Non-localhost callbacks remain supported separately through `RemoteRedirectURI`; this change does not modify that flow.


EthanHeilman:
In OPKSSH we added the ability to use non-localhost redirectURI to support bastionhosts that don't run on the user's machine.  Not sure if it conflicts with this PR or not but worth looking over

https://github.com/openpubkey/opkssh/pull/456

sukujgrg:
I am not seeing a reason it conflicts because with this change, it is honouring what is defined in the client side regardless whether it is configured to use localhost or 127.0.0.1 or some domain name.
I will look more into this opkssh#456 and try to make more informed judegement.

sukujgrg:
I looked through [opkssh#456](https://github.com/openpubkey/opkssh/pull/456) in more detail. It does not conflict with this PR.

OPKSSH’s non-localhost callback is handled separately through `RemoteRedirectURI`. That value overrides the URI sent to the OIDC provider, while `RedirectURIs` still identifies the local listener to which the remote service proxies the callback. This PR does not modify the `RemoteRedirectURI` path.

As a clarification to my previous comment, this PR does not allow an arbitrary domain in `RedirectURIs`; it requires `localhost` or a loopback IP. My earlier “some domain name” phrasing was incorrect.

EthanHeilman (approved):
LGTM

sukujgrg:
@EthanHeilman  Merge conflict resolved. CI is now passing - no functional changes since approval. Thanks!

EthanHeilman (approved):
LGTM, thanks for your work on this. It is much cleaner now

PR #377, opened 2026-07-13
refactor: remove direct logrus usage

sukujgrg:
This MR explores removing OpenPubkey’s direct Logrus usage and moving authorization URL presentation to the client application. I’m not sure whether the preferred design is for the library to continue opening/printing the URL or to leave that responsibility to callers, so feedback on this API direction is welcome. The current patch preserves automatic opening and the existing manual fallback for compatibility.

# Summary

Remove OpenPubkey’s direct Logrus usage and let client applications control how authorization URLs are presented or logged.

## Motivation

Running the simple example on the current `main` branch produces library-owned Logrus output:

```text
INFO[0000] listening on http://127.0.0.1:3000/
INFO[0000] press ctrl+c to stop
INFO[0000] Opening browser to http://localhost:3000/login
```

Because OpenPubkey is primarily consumed as a Go library, so emitting informational messages through a global logging framework prevents client applications from fully controlling their logging and user-facing output. This change removes those messages while preserving automatic browser opening. Applications that need the authorization URL can receive it through `AuthorizationURLHandler`.

## Changes

- Remove direct Logrus imports and calls.
- Return authorization handler and callback-server failures to callers.
- Add a framework-neutral `AuthorizationURLHandler` for providers and `WebChooser`.
- Invoke the handler before automatic browser opening so applications receive the URL if opening fails.
- Keep `SetOpenBrowserOverride` as a deprecated, compatible alias.
- Preserve the manual-browser stderr fallback when automatic opening is disabled or fails.
- Update the README and simple example.

Client applications require no changes for the default automatic-browser flow. Applications that want to print, log, or present the URL can configure:

```go
err := providers.SetAuthorizationURLHandler(op, func(url string) error {
	fmt.Println("Authentication URL:", url)
	return nil
})
```

The handler receives the URL with automatic browser opening either enabled or disabled.

Logrus remains an indirect dependency through `zitadel/oidc`, but OpenPubkey no longer imports or uses it directly.

**Note:**
Zitadel is separately tracking removal of its logging-library dependency in [https://github.com/zitadel/oidc/issues/829](https://github.com/zitadel/oidc/issues/829), which should allow the remaining indirect Logrus dependency to disappear after a future dependency update.

## Validation

- `go test ./...`
- `go vet ./...`
- `git diff --check`


client/choosers/web_chooser.go
```diff
@@ -220,23 +242,30 @@ func (wc *WebChooser) ChooseOp(ctx context.Context) (providers.OpenIdProvider, e
...
 		}
 
-		if wc.OpenBrowser {
-			logrus.Infof("Opening browser to %s", loginURI)
-			if err := util.OpenUrl(loginURI); err != nil {
-				logrus.Errorf("Failed to open url: %v", err)
+		if wc.authorizationURLHandler != nil {
+			if err := wc.authorizationURLHandler(loginURI); err != nil {
+				_ = wc.server.Shutdown(ctx)
+				return nil, fmt.Errorf("authorization URL handler failed: %w", err)
```
EthanHeilman: "Failed to open url:" is a much more clear message to the user than "authorization URL handler failed:"

I have a distant memory of  returning error here being problematic. IIRC the issue is on some system configurations the OS will make an assessment if an application should be allowed to open a URI. Its better to just display a error, but continue so the user still has the option to copy paste the URI into their browser.

client/choosers/mock_web_chooser.go
```diff
@@ -54,8 +53,6 @@ func BrowserOpenOverride(opToChoose string) func(string) error {
 			// Use retry logic here as well to handle timing issues with the OP server
 			resp, err := retryHTTPGet(selectOpUri)
 			if err != nil {
```
EthanHeilman, line 61: Removing logus here is great but we should still capture the error. Maybe add OutWriter io.Writer on the struct and have it default to writing to stderr.

client/choosers/web_chooser.go
```diff
@@ -50,16 +50,32 @@ var chooserTemplateFile string
...
-	server              *http.Server
-	browserOpenOverride BrowserOpenOverrideFunc
+	OpList                  []providers.BrowserOpenIdProvider
+	opSelected              providers.BrowserOpenIdProvider
+	OpenBrowser             bool
+	useMockServer           bool
+	mockServer              *httptest.Server
+	server                  *http.Server
+	authorizationURLHandler AuthorizationURLHandler
+	openBrowser             func(string) error
```
EthanHeilman: `OpenBrowser` being a boolean but `openBrowser` being a func is a bit confusing. How does `browserOpenOverride` differ from `openBrowser`?

client/choosers/web_chooser.go
```diff
@@ -220,23 +242,30 @@ func (wc *WebChooser) ChooseOp(ctx context.Context) (providers.OpenIdProvider, e
...
-		if wc.browserOpenOverride != nil {
-			if err := wc.browserOpenOverride(loginURI); err != nil {
-				logrus.Errorf("Failed to open url: %v", err)
-				return nil, err
+		if wc.OpenBrowser {
+			openBrowser := wc.openBrowser
+			if openBrowser == nil {
+				openBrowser = util.OpenUrl
 			}
+			if err := openBrowser(loginURI); err != nil && wc.authorizationURLHandler == nil {
```
EthanHeilman: I might be missing something here, what's the difference between wc.openBrowser and authorizationURLHandler? they seem like do the same thing?

EthanHeilman:
Removing logging here is something the project has wanted for a while. 

>  I’m not sure whether the preferred design is for the library to continue opening/printing the URL or to leave that responsibility to callers, so feedback on this API direction is welcome.

Ideally we would provide a interface that the calling library can plug into stdout or into a logger.

opkssh has a writer in the struct for login, if we supported that here, opkssh just pass the OutWriter to the client and the client could set it to the providers, web_chooser. Something similar could work for verify.

```
OutWriter io.Writer // Captures non-logged output that would normally be written to stdout
```

client/choosers/web_chooser.go
EthanHeilman: Is authorizationURLHandler handling the authorization URI? Would it be the web chooser URI?

client/choosers/web_chooser.go
```diff
@@ -220,23 +242,30 @@ func (wc *WebChooser) ChooseOp(ctx context.Context) (providers.OpenIdProvider, e
...
 
-		if wc.browserOpenOverride != nil {
-			if err := wc.browserOpenOverride(loginURI); err != nil {
-				logrus.Errorf("Failed to open url: %v", err)
-				return nil, err
+		if wc.OpenBrowser {
+			openBrowser := wc.openBrowser
+			if openBrowser == nil {
+				openBrowser = util.OpenUrl
 			}
```
EthanHeilman: I think I get this. Previously we'd set OpenBrowser = false and then define browserOpenOverride, but now we can set OpenBrowser = true and set openBrowser as the override and you won't see the browser open errors messages in tests?

EthanHeilman (commented):
Trying to understand the intent here

client/choosers/web_chooser.go
EthanHeilman: Wait is authorizationURLHandler only intended to capture log messages?

EthanHeilman: Oh, ok, I get it now. Sorry I've been reading code all day and my brain is fired. `authorizationURLHandler` is used so that someone calling this as a library can extract the URI and take some action on it.

Do we need browserOverride/openBrowser for the tests? Couldn't we just set openBrowser=false and then use authorizationURLHandler do what browserOverride was doing before?

client/choosers/web_chooser.go
EthanHeilman: Ahhh! Because authorizationURLHandler isn't intended to actually open the URL, but just capture the URL. An error here is unexpected should fail! Resolve this comment after reading.

client/choosers/web_chooser.go
sukujgrg: Yes, the existing mock interaction tests can set `OpenBrowser=false` and use `AuthorizationURLHandler`; they don’t need a separate private browser-opener hook. Deprecated `SetOpenBrowserOverride` remains for source compatibility. The private `openBrowser` field is different: it injects the real browser-opening operation so we can deterministically test the browser-open failure path. I’ll rename to `browserOpener` to make that distinction clearer.

client/choosers/web_chooser.go
sukujgrg: Not quite. Mocks use `OpenBrowser=false` with `AuthorizationURLHandler`. I’ll rename the private `openBrowser` to `browserOpener`; and it only lets tests simulate browser-opening failures.

client/choosers/web_chooser.go
sukujgrg: agreed, now naming feels confusing. `OpenBrowser` controls whether automatic opening is attempted, while the private `openBrowser` (I will rename to `browserOpener`) is the actual opener used for failure testing. `browserOpenOverride` is the deprecated URL callback.

EthanHeilman:
This looks good, but I want to read it again when my brain is fresh. I also want to test it against opkssh. I'll probably approve and merge on Sunday,

client/choosers/web_chooser.go
```diff
@@ -50,16 +51,37 @@ var chooserTemplateFile string
...
+// The handler is called before automatic browser opening is attempted. This
+// ensures the application receives the URL even if the browser cannot be
+// opened. When automatic browser opening is disabled, the application can use
+// the handler to display the URL, open a browser, or integrate the URL into its
+// own user interface. Returning an error aborts provider selection and is
+// returned by WebChooser.ChooseOp. If automatic browser opening fails after a
+// handler has received the URL, the opening error is not returned; the handler
+// is expected to have made the URL available through an application-controlled
+// mechanism.
+type AuthorizationURLHandler func(url string) error
```
EthanHeilman: Rename this to `BeforeBrowserOpenURIHook`

- It is not always an AuthorizationURL, for instance in webchooser is the webchooser URI.
- In the OPKSSH codebase we referred to functions like this as hooks and not handlers.

client/choosers/web_chooser.go
```diff
@@ -50,16 +51,37 @@ var chooserTemplateFile string
...
+	OpList                  []providers.BrowserOpenIdProvider
+	opSelected              providers.BrowserOpenIdProvider
+	OpenBrowser             bool
+	useMockServer           bool
+	mockServer              *httptest.Server
+	server                  *http.Server
+	authorizationURLHandler AuthorizationURLHandler
+	// outWriter receives non-fatal, user-facing messages. If nil, os.Stderr is
+	// used for backward compatibility.
+	outWriter io.Writer
```
EthanHeilman: Nit: if outWriter is intended only for error messages, rename to errWriter. We might want to add an outWriter for stdout at some point.

EthanHeilman: Actually we need an outWriter (stdout) and errWriter (errout) since on [line 268](https://github.com/openpubkey/openpubkey/pull/377/changes#diff-45d886d89330a9b6d96ac6301c031e0aeb53d864a5181e046eee574e93850cc1R268) we use the outWriter for a message which isn't an error message:

```golang
_, _ = fmt.Fprintf(wc.outputWriter(), "Open your browser to: %s\n", loginURI)
```

client/choosers/web_chooser.go
```diff
@@ -220,23 +248,33 @@ func (wc *WebChooser) ChooseOp(ctx context.Context) (providers.OpenIdProvider, e
...
+			browserOpener := wc.browserOpener
+			if browserOpener == nil {
+				browserOpener = util.OpenUrl
+			}
+			if err := browserOpener(loginURI); err != nil {
+				_, _ = fmt.Fprintf(wc.outputWriter(), "Failed to open URL: %v\n", err)
+				if wc.authorizationURLHandler == nil {
+					// Preserve the previous continue-on-failure behavior while ensuring
+					// the user can still complete authentication manually.
+					_, _ = fmt.Fprintf(wc.outputWriter(), "Open your browser to: %s\n", loginURI)
```
EthanHeilman: Since we print this out in most the if and the else condition, we don't need to duplicate it. Not sure we even need the `if wc.authorizationURLHandler == nil` since the outWriter allows the caller to determine what happens to that message.

```
// Preserve the manual-browser behavior for existing applications. New
// applications should install an AuthorizationURLHandler and decide how
// to present the URL.
_, _ = fmt.Fprintf(wc.outputWriter(), "Open your browser to: %s\n", loginURI)
```

providers/standard_provider.go
```diff
@@ -280,9 +284,7 @@ func (s *StandardOp) defaultRequestTokens(ctx context.Context, cicHash string) (
 	}
 
 	shutdownServer := func() {
-		if err := s.server.Shutdown(ctx); err != nil {
-			logrus.Errorf("Failed to shutdown http server: %v", err)
-		}
+		_ = s.server.Shutdown(ctx)
```
EthanHeilman: We should write this to errWriter as it makes debugging the server shutdown bugs much easier. Ask me how I know =)

providers/standard_provider.go
```diff
@@ -339,26 +342,35 @@ func (s *StandardOp) defaultRequestTokens(ctx context.Context, cicHash string) (
...
-	// This is to mock out the OP interactions.
-	if s.browserOpenOverride != nil {
-		if err := s.browserOpenOverride(loginURI); err != nil {
-			logrus.Errorf("Failed to open url to mock browser: %v", err)
+		if err := browserOpener(loginURI); err != nil {
+			_, _ = fmt.Fprintf(s.outputWriter(), "Failed to open URL: %v\n", err)
+			if s.authorizationURLHandler == nil {
+				// Preserve the previous continue-on-failure behavior while ensuring
+				// the user can still complete authentication manually.
+				_, _ = fmt.Fprintf(s.outputWriter(), "Open your browser to: %s\n", loginURI)
```
EthanHeilman: We should use the outWriter, errWriter pattern I proposed for webchooser here as well. If the provider is inside of a webchooser, the webchooser should default the provider to the same writers as the webchooser, unless the provider has a specifically set chooser.

providers/standard_provider.go
```diff
@@ -339,26 +342,35 @@ func (s *StandardOp) defaultRequestTokens(ctx context.Context, cicHash string) (
...
+			}
 		}
+	} else if s.authorizationURLHandler == nil {
+		// Preserve the manual-browser behavior for existing applications. New
```
EthanHeilman: We don't need to duplicate this

providers/standard_provider.go
```diff
@@ -460,10 +469,7 @@ func (s *StandardOp) deviceFlowRequestTokens(ctx context.Context, cicHash string
 	}
 
 	code, err := qrcode.Create(qrCodeURL)
-	if err != nil {
-		// do not fail, just log the error, user can still manually open the url
-		logrus.Warnf("could not create qrcode, fallback to textual representation only: %s", err)
-	} else {
+	if err == nil {
```
EthanHeilman: Don't silently fail here. OutWriter, ErrWriter pattern

EthanHeilman (changes requested):

client/choosers/web_chooser.go
sukujgrg: Done

providers/standard_provider.go
sukujgrg: Done. Writing to stderr now via new errorWriter.

providers/standard_provider.go
sukujgrg: Done.

client/choosers/web_chooser.go
sukujgrg: Agreed. Deduplicated into `presentChooserURI` and removed the hook-nil check. We still print only when manual navigation is needed; callers that want to suppress that message can set `OutWriter` to `io.Discard`.

EthanHeilman (approved):
LGTM, I made some minor changes, renamed some variables and refactored the writers so deduplicate some code

PR #379, opened 2026-07-14
[fix] Changes GQ signature to cover GQ protected header 

kipz:
## Summary
- Add a test proving the GQ signature does not cover the outer protected header
- Fix GQ sign/verify to bind the outer protected header into the signature

🤖 Generated with [Claude Code](https://claude.com/claude-code) - Reviewed and edited by @EthanHeilman 

This moves the GQ signature from signing the JWT payload and protected headers, to signing the JWT payload and GQ protected headers. The GQ signature still covers the JWT protected headers because the JWT protected headers are stored in the GQ protected header as the `QG_JWT.protected_headers.kid=base64(JWT.protected_headers)`

## Security Impact

This only has a security impact for GitLab GQ-commitment PK Tokens. GitHub only uses GQ signatures for replay protection and binds the users public key to the ID Token via the audience (`aud`) claim.

See [Docs/PK Tokens - GQ-Commitment PK Tokens (Gitlab-CI Example)](https://github.com/openpubkey/openpubkey/blob/d30b081b83f60f70793b124e46ffcd2620286b4d/docs/pktoken.md#gq-commitment-pk-tokens-gitlab-ci-example)

## Compatibility

This is a non-backwards compatible change:
  - Signatures produced by code prior to this change will not verify on code after this change.
  - Signatures produced by code after this change will not verify on code prior to this change.

### DIFF
--- a/gq/sign.go
+++ b/gq/sign.go
@@ -106,8 +106,6 @@ func (sv *signerVerifier) SignJWT(jwt []byte, opts ...Opts) ([]byte, error) {
 		return nil, err
 	}
 
-	signingPayload := util.JoinJWTSegments(origHeaders, payload)
-
 	headers := jws.NewHeaders()
 	err = headers.Set(jws.AlgorithmKey, jose.GQ256)
 	if err != nil {
@@ -150,7 +148,9 @@ func (sv *signerVerifier) SignJWT(jwt []byte, opts ...Opts) ([]byte, error) {
 
 	defer private.Destroy()
 
-	gqSig, err := sv.Sign(private.Bytes(), signingPayload)
+	// Sign over the new outer header so jkt/cic/extra claims are bound.
+	gqMessage := util.JoinJWTSegments(headersEnc, payload)
+	gqSig, err := sv.Sign(private.Bytes(), gqMessage)
 	if err != nil {
 		return nil, err
 	}
--- a/gq/sign.go
+++ b/gq/sign.go
@@ -148,7 +148,7 @@ func (sv *signerVerifier) SignJWT(jwt []byte, opts ...Opts) ([]byte, error) {
 
 	defer private.Destroy()
 
-	// Sign over the new outer header so jkt/cic/extra claims are bound.
+	// Sign over the new GQ protected header so jkt/cic/extra claims are covered by the GQ signature and can not be altered without breaking signature verification
 	gqMessage := util.JoinJWTSegments(headersEnc, payload)
 	gqSig, err := sv.Sign(private.Bytes(), gqMessage)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


providers/providerverifier.go
```diff
@@ -236,7 +236,7 @@ func (v *DefaultProviderVerifier) verifyCommitment(idt *oidc.Jwt, cic *clientins
 				v.options.GQAudiencePrefix, audStr)
 		}
 
-		// Get the commitment from the GQ signed protected header claim "cic" in the ID Token
+		// "cic" is in the outer header, which the GQ signature binds.
```
EthanHeilman: I don't think this comment change is needed

```suggestion
		// Get the commitment from the GQ signed protected header claim "cic" in the ID Token
```

gq/verify.go
```diff
@@ -96,14 +96,17 @@ func (sv *signerVerifier) VerifyJWT(jwt []byte) bool {
...
 	}
 
-	_, payload, signature, err := jws.SplitCompact(jwt)
+	gqHeaders, payload, signature, err := jws.SplitCompact(jwt)
 	if err != nil {
 		return false
 	}
 
-	signingPayload := util.JoinJWTSegments(origHeaders, payload)
+	// identity (from kid) forms G and must match the signer's RSA key;
```
EthanHeilman: This comment is misleading as the varible below is named identity and isn't not just the kid. Kid does not form G.
```suggestion
```

gq/verify.go
```diff
@@ -96,14 +96,17 @@ func (sv *signerVerifier) VerifyJWT(jwt []byte) bool {
...
+	// message covers the outer header so tampering it fails verification.
```
EthanHeilman:
```suggestion
	// This ensures GQ signature covers the GQ protected header so an attacker can not alter the values in the GQ protected header
```

gq/sign.go
```diff
@@ -150,7 +148,9 @@ func (sv *signerVerifier) SignJWT(jwt []byte, opts ...Opts) ([]byte, error) {
 
 	defer private.Destroy()
 
-	gqSig, err := sv.Sign(private.Bytes(), signingPayload)
+	// Sign over the new outer header so jkt/cic/extra claims are bound.
```
EthanHeilman:
```suggestion
	// Sign over the new GQ protected header so jkt/cic/extra claims are covered by the GQ signature and can not be altered without breaking signature verification
```

EthanHeilman (changes requested):

EthanHeilman (approved):

EthanHeilman:
This is a non-backwards compatible change:
- Signatures produced by code prior to this change will not verify on code after this change.
- Signatures produced by code after this change will not verify on code prior to this change.

EthanHeilman (approved):
LGTM

PR #386, opened 2026-07-18
feat(providers): add Forgejo Actions OpenID Provider

Blacks-Army:
Adds `ForgejoOp`, a provider for Forgejo Actions (e.g. Codeberg), modelled on `GithubOp`. Forgejo runners inject the same `ACTIONS_ID_TOKEN_REQUEST_URL` and `ACTIONS_ID_TOKEN_REQUEST_TOKEN` variables as GitHub, but the issuer is per instance (`<instance>/api/actions`), so it is derived from the token request URL instead of being fixed.

New exported API in `providers`:
- `NewForgejoOp(issuer, tokenURL, token)`
- `NewForgejoOpFromEnvironment()`
- `IsForgejoIssuer(issuer)`
- `ForgejoIssuerFromTokenRequestURL(url)`

Tokens are GQ-signed and verified with an audience commitment, the same flow `GithubOp` uses. Tests cover issuer/env parsing and a request/verify roundtrip.

Enables Forgejo/Codeberg support in opkssh (openpubkey/opkssh#519).


providers/forgejo_actions.go
```diff
@@ -0,0 +1,208 @@
+// Copyright 2025 OpenPubkey
```
EthanHeilman:
```suggestion
// Copyright 2026 OpenPubkey
```

Blacks-Army: Updated the year, it was a copy paste mistake.

Blacks-Army:
@EthanHeilman FYI: I didn’t want to modify the GitHub provider, so I created a standalone Forgejo provider instead. It could probably be refactored or consolidated later, though.

EthanHeilman:
@Blacks-Army This looks pretty good, two things:

- Can you link to an example project using this and working?
- Add a Forgejo Actions ID Token to the ID Token zoo here https://github.com/openpubkey/openpubkey/blob/main/docs/idtokens.md

Blacks-Army:
Thanks @EthanHeilman !

For the example project, do you mean a repo on Codeberg using this? I can push one of my local repos there and add a workflow that SSHs into a local SSH server if that’s enough.

For the ID Token, do you mean adding a Forgejo Actions ID Token like the one described here? https://forgejo.org/docs/latest/user/actions/security-openid-connect/ 

I couldn’t find a GitHub ID Token example either, so is one actually required?

Blacks-Army:
@EthanHeilman 

I've added Forgejo Actions example entries to the docs, an ID Token in `docs/idtokens.md` and the JWKS in `docs/jwks.md`. 

And here's a working example on Codeberg with a passing run, [repo](https://codeberg.org/Blacks-Army/opkssh) and [run](https://codeberg.org/Blacks-Army/opkssh/actions/runs/3/jobs/0/attempt/1), that builds opkssh, logs in via `opkssh login forgejo`, and SSHes into localhost with the resulting certificate.

EthanHeilman:
@Blacks-Army 

> For the ID Token, do you mean adding a Forgejo Actions ID Token like the one described here? https://forgejo.org/docs/latest/user/actions/security-openid-connect/

Yeah

> I couldn’t find a GitHub ID Token example either, so is one actually required?

Well that's a mistake, we should have a  GitHub ID Token. It isn't required, I just want it there for future reference when debugging stuff.

Blacks-Army:
> @Blacks-Army 
> 
> 
> 
> > For the ID Token, do you mean adding a Forgejo Actions ID Token like the one described here? https://forgejo.org/docs/latest/user/actions/security-openid-connect/
> 
> 
> 
> Yeah
> 
> 
> 
> > I couldn’t find a GitHub ID Token example either, so is one actually required?
> 
> 
> 
> Well that's a mistake, we should have a  GitHub ID Token. It isn't required, I just want it there for future reference when debugging stuff.

@EthanHeilman 

See my latest comment. The JWKS and ID tokens are different for each Forgejo instance. The examples I added are from Codeberg. The kid at least.

EthanHeilman (approved):

PR #388, opened 2026-07-22
feat: add timeouts to browser authentication flows

sukujgrg:
## Summary

Add a configurable timeout to browser authentication flows, defaulting to 10 minutes.

This applies to the web chooser, OIDC callback, and cosigner callback servers. It also ensures library-owned servers close cleanly on success, failure, cancellation, or timeout while preserving caller context errors.

A negative timeout disables the library deadline. Device flow is unchanged.

## Compatibility

Clients using constructors or keyed struct literals are unaffected. `opkssh` compiles against these changes without modification.

## Testing

- Verified `opkssh` compile against the updated library

### DIFF
--- a/client/choosers/web_chooser.go
+++ b/client/choosers/web_chooser.go
@@ -28,7 +28,9 @@ import (
 	"net/http/httptest"
 	"sort"
 	"strings"
+	"time"
 
+	"github.com/openpubkey/openpubkey/internal/httpserver"
 	"github.com/openpubkey/openpubkey/providers"
 	"github.com/openpubkey/openpubkey/util"
 )
@@ -50,13 +52,17 @@ var chooserTemplateFile string
 
 type WebChooser struct {
 	util.OutOrErrWriter
-	OpList        []providers.BrowserOpenIdProvider
-	opSelected    providers.BrowserOpenIdProvider
-	OpenBrowser   bool
-	useMockServer bool
-	mockServer    *httptest.Server
-	server        *http.Server
-	loginURIHook  LoginURIHook
+	OpList      []providers.BrowserOpenIdProvider
+	opSelected  providers.BrowserOpenIdProvider
+	OpenBrowser bool
+	// AuthWaitTimeout bounds how long ChooseOp waits for the user to pick a
+	// provider in the browser. Zero means providers.DefaultAuthWaitTimeout
+	// (10 minutes). A negative value disables the library timeout so only the
+	// parent context can cancel the wait.
+	AuthWaitTimeout time.Duration
+	useMockServer   bool
+	server          *http.Server
+	loginURIHook    LoginURIHook
 	// browserOpener replaces util.OpenUrl in tests that exercise browser-open
 	// success and failure paths.
 	browserOpener func(string) error
@@ -117,6 +123,10 @@ func (wc *WebChooser) ChooseOp(ctx context.Context) (providers.OpenIdProvider, e
 	if err != nil {
 		return nil, err
 	}
+	authCtx, cancelAuth := providers.WithAuthWaitTimeout(ctx, wc.AuthWaitTimeout)
+	chooseDone := make(chan struct{})
+	defer close(chooseDone)
+	shutdownCh := make(chan struct{}, 1)
 
 	mux.HandleFunc("/chooser", func(w http.ResponseWriter, r *http.Request) {
 		type Provider struct {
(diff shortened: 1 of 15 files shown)


sukujgrg:
@EthanHeilman, I’d appreciate your input on the default policy before moving this out of draft. Many CLI tools bound interactive browser authentication so an abandoned flow doesn’t leave a callback server running indefinitely. This PR follows that pattern with a configurable 10-minute default while still allowing callers to disable it or provide an earlier context deadline. Does that behavior align with how you’d like OpenPubkey to handle abandoned authentication flows?

EthanHeilman:
@sukujgrg A timeout on login seems reasonable. There is a question if the openpubkey library should handle this or if the code calling openpubkey should handle it.


For instance, something like this
```
...
if opkClient, err := client.New(op); err != nil {
  return err
}

timeout := 10*time.minute
ctx, cancel := context.WithTimeout(context.Background(), timeout)
defer cancel()

pkt, err := opkClient.Auth(ctx)
```

It doesn't unreasonable to include this as a feature in OpenPubkey itself, but it should be part of the client, not the providers. That should be a 7 line change.

sukujgrg:
Thanks - callers can already wrap Auth with `context.WithTimeout`, and this PR continues to honor that. However, that alone doesn’t ensure OpenPubkey’s localhost servers are cleaned up when the context ends. Currently, WebChooser can return on `ctx.Done()` without closing its server, and the OIDC session-hook path can also remain running.
Since these server-owning APIs can be used independently of `client.Auth`, this PR handles the timeout and cleanup where each server is created. The default remains configurable and can be disabled (`AuthWaitTimeout: -1`) if the caller wants full control. Would that approach be acceptable?

EthanHeilman:
>Since these server-owning APIs can be used independently of client.Auth, this PR handles the timeout and cleanup where each server is created.

I might be misunderstanding you. Are you saying to use providers without the client? What's the usecase hedre?

> Currently, WebChooser can return on ctx.Done() without closing its server, and the OIDC session-hook path can also remain running.

That seems like a bug then and independent of timeouts. If WebChooser is done, it should close its server, no?

I understand why someone might want client.Auth to have a timeout, but I don't understand what someone is doing to need webchooser by itself to have a timeout.

sukujgrg:
You're right that cleanup on cancellation is a separate bug, and `ChooseOp` should close its server when the wait ends without a successful selection.

The WebChooser case isn't about using providers without the client. For example, the existing `./examples/userinfo` flow starts the chooser before constructing the client:

```go
op, err := choosers.NewWebChooser(...).ChooseOp(ctx)
opkClient, err := client.New(op)
pkt, err := opkClient.Auth(ctx)
```

If the browser fails to open or the user abandons the chooser, execution remains blocked in `ChooseOp`:

```text
Opening browser to http://localhost:54442/chooser
Failed to open URL: exec: "open": executable file not found in $PATH
Open your browser to: http://localhost:54442/chooser
```

Without a deadline, that listener stays active until the caller cancels the context or the process exits, and `client.Auth` is never reached. A timeout added only inside `client.Auth` therefore wouldn't cover this part of the normal login flow.

The same abandoned-wait behavior can occur during the OIDC callback in `./examples/simple`:

```text
Authorization Request URI: http://localhost:3000/login
Opening browser to http://localhost:3000/login
Failed to open URL: exec: "open": executable file not found in $PATH
Open your browser to: http://localhost:3000/login
```

The successful paths already clean up these listeners. The gap is when the user abandons the flow or the wait ends through failure or cancellation. Since OpenPubkey creates and owns the listeners, it should close them in those cases as well. The default deadline gives an otherwise-abandoned flow an endpoint, while callers can disable it and manage the deadline themselves.

My intent with this draft is to bound the chooser and OIDC/cosigner callback waits, ensure each server is cleaned up when the flow ends, and still honor an earlier caller deadline.

EthanHeilman:
> If the browser fails to open or the user abandons the chooser, execution remains blocked in ChooseOp:

Good point. I forgot how this worked, I thought the chooser would wait until client.Auth took an action on the op and the webChooser would then start the server. ChooseOp should really be called inside the client so the client can manage it. Everything that happens outside the client should be configuration.

The client should be fully responsible for starting and stopping servers. What changes would need to happen to make this true?

sukujgrg:
Moving provider selection into the client would require the client to accept chooser/configuration rather than an already-selected OP, invoke ChooseOp(ctx) from Auth, and update the chooser-based examples such as userinfo, web, and deviceflow. We would likely need a new constructor or option to preserve compatibility with callers that already pass a concrete OP.

Even with that change, the server-owning code still needs to close its listeners when the shared context gets canceled. This PR adds that cleanup for the three localhost servers involved in browser authentication: the web chooser, OIDC callback, and cosigner callback. It also gives each wait its own configurable 10-minute default; a negative value disables that library deadline, and an earlier caller deadline still wins.

My preference is to keep this PR scoped to cleanup and abandoned flows, then handle the chooser-in-client API change separately. I’m marking this ready for review unless you’d prefer both changes together.

sukujgrg:
To clarify “cleanup”: on main, ChooseOp can return on cancellation without closing its localhost listener, and its redirect handler can remain blocked indefinitely.

This PR closes the server on cancellation, timeout, error, and completed redirect; makes handler waits cancellation-aware; and uses bounded shutdown with forced close as a fallback. Moving ChooseOp into the client could centralize timeout policy, but this server-lifecycle cleanup would still be required.

PR #392, opened 2026-07-29
chore(deps): upgrade zitadel/oidc from v3 to v4

EthanHeilman:
Zitadel v4 isn't stable yet, so we should not merge this until we have a stable release.

Update all import paths from github.com/zitadel/oidc/v3 to v4. No call site changes were required.

Decode userinfo responses into a new rawUserInfo type rather than oidc.UserInfo. v4 promotes email_verified and phone_number_verified to oidc.Bool, whose UnmarshalJSON errors on any value that is not a boolean or the strings "true"/"false", so a provider sending "email_verified": null or 1 would fail the entire userinfo request instead of returning the remaining claims. rawUserInfo implements rp.SubjectGetter, so sub verification against the ID token is unchanged, and it returns the provider's response body verbatim.

This does change one behavior: a string "true" is no longer normalized to boolean true, because the body is now passed through exactly as the OP sent it.


kipz (approved):
LGTM

kipz:
There is a `v4.0.0-next.4` release that might be worth trying....

EthanHeilman:
@kipz Thanks
 for the review. I'm waiting until they have an official release. I got confused between zitadel v4, which has been out for a while and zitadel/oidc v4 which isn't out yet.

PR #395, opened 2026-08-12
feat(providers): OIDC key binding for the authorization-code and refresh flows

EthanHeilman:
Consume zitadel/oidc's native OpenID Connect Key Binding API (rp.WithKeyBinding) for the code and refresh flows, replacing the transport-based DPoP shim. The device authorization flow is out of scope here and fails closed with a clear error if combined with key binding; device support lands in a follow-up.

Includes a TEMP go.mod replace to the zitadel/oidc kb-support branch until it is released upstream.


kipz (approved):
Nice! :)

PR #396, opened 2026-08-14
Default StandardOp to be refreshable

EthanHeilman:
OpenPubkey has long maintained `StandardOp` and `StandardOpRefreshable` types so that we can infer based on the type if an OP can be refreshed. Since these types were hidden behind the interface `OpenIdProvider` this did provide compile benefits and we need to check the type at run time with checks like:

```golang
func (o *OpkClient) Refresh(ctx context.Context) (*pktoken.PKToken, error) {
	if tokensOp, ok := o.Op.(providers.RefreshableOpenIdProvider); ok {
		if o.refreshToken == nil {
			return nil, fmt.Errorf("no refresh token set")
		}
```

This PR changes this so that `StandardOp` is Refreshable. Rather than relying on the type, we check if a refresh token was provided and if Refresh is called and their is not Refresh Token we fail. The scope `offline` for requesting a RefreshToken is still configurable. `StandardOpRefreshable` is now just a alias to `StandardOp`

The reason for this change is that almost every OP is refreshable and when people use custom providers with the `StandardOp` we should just default to these OPs being refreshable. The caller already has to decide to call refresh or not.

Adds a `HasRefreshToken` method on client to allow callers to check if refreshing is supported.


kipz:
LGTM!

